PR #92, opened 2012-08-07
The git repo of ctags now supports context and describes for ruby

sundbp:
So I've added in those types. It doesn't yet display nicely, but I think that's more to do with the state of the ctags support as with the tagbar parsing of the output.

At least with this in there we don't get lots of error messages.


sundbp:
I've forgotten what this is about I'm afraid. I'm no longer using it
personally. It is likely pragmatic to delete/close the PR.

alerque:
Thanks for the reply @sundbp, and you might be right. Since some other people were interested in this at some point I think I'll leave it open for just a little bit to solicit feedback. If I get through the contribution backlog and this is still around it will be a candidate for closing as no longer relevant. Still thanks for attempting the contribution!

PR #192, opened 2014-01-30
Better support for clojure.

kul:
Adding better support for clojure. Here is reference ctags file
https://gist.github.com/kul/8704283


tfoldi:
You don't need the additional `ctags` file as Universal Ctags has native clojure support. It's pretty bad actually, but works so I would need this PR merged as well

PR #238, opened 2014-12-20
add '-' flag before static functions in c

Cofyc:
I did this for my own needs, and only tested with c source files (ctags: 5.8, vim: 7.4). It may break in other situations, but I propose this feature (maybe implemented in other ways).

When there is a lot of functions in source file, I think it's convenient that we can distinct static functions and functions that may be used by outside.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1266,6 +1266,11 @@ function! s:BaseTag._getPrefix() abort dict
         endif
     endif
 
+    " File-restricted scoping
+    if has_key(self.fields, 'file')
+        let prefix .= '-'
+    end
+
     return prefix
 endfunction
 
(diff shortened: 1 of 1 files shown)


majutsushi:
Hey, sorry for the late reply. The code looks code, but I noticed that if I run ctags on a C file with the file scope enabled then it reports a lot of tags as having file scope, not just the ones that are declared static. Does that happen for you too? I'm not sure the scope information would really be useful if it is going to be unreliable...

Cofyc:
It works for me.

screen shot 2015-01-31 at 7 35 42 pm

This is my c test file, and I've used this feature a lot in my c projects, can you show me your c test file?

```c
#include <stdio.h>

static void
static_func()
{
    printf("hello world\n");
}

int
non_static_func()
{
    return 0;
}

int
main(void)
{
    static_func();
    non_static_func();
}
```

```
$ ctags --version                                            
  Compiled: Nov 19 2013, 10:33:30
```

```
$ vim --version
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Dec 15 2014 11:27:33)
MacOS X (unix) version
Included patches: 1-488
```

majutsushi:
You're right, I got a bit confused with regard to structs. I have now merged your code together with a small fix, thanks!

PR #264, opened 2015-04-24
Fix vim-airline: Suppress status-line output when executing shell

h3xx:
Failing to suppress output can cause problems in some environments,
especially if the shell command fails or does something else spooky.

Example where failing to suppress the shell command causes issues with
[vim-airline](https://github.com/bling/vim-airline) with the tagbar
extension enabled:

screenshot1

Here's another example:

screenshot2

Both were taken using uxterm in Linux.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3665,7 +3665,7 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call s:debug(v:statusmsg)
         redraw!
     else
-        let ctags_output = system(a:ctags_cmd)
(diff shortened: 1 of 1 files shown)


h3xx:
I debugged why this was happening for two hours and finally figured out it happened ONLY when called from within Airline, and mostly when the file (or empty buffer) is first opened. Good idea anyway?

majutsushi:
Huh, that's some pretty strange output. I've never seen that before, probably because I don't use Airline. Thanks for debugging it and the patch!

PR #273, opened 2015-06-05
Add support for Universal Ctags

vhda:


majutsushi:
Thanks! I've committed a slightly different version now.

PR #278, opened 2015-06-23
Perl Moose support for types and attributes

imran-uk:
- requires additional config for ctags


imran-uk:
Wow this is from 4 years ago!

OK, I will try a rebase and resubmit the pull-request. Never done that
before but I'll figure it out :)

alerque:
I can help with specific git stuff if you need help. You don't have to resubmit, you just have to push to the same branch on github (in this case with a force) and the PR will update.

alerque:
I just tried to rebase this on master to get the PR out of the way and it looks like this has been basically obsoleted by updates to the Universal Ctags support. Specifically the Perl6 type has support for all these tags, and using UCtags doesn't require any `ctags` preferences changes on the users end.

I think I'm going to close this with the recommendation that people just use Universal Ctags and get support out of the box. If legacy support is needed or the types need to be adjusted in Perl instead of Perl6 I'm happy to accept a PR along those lines.

PR #280, opened 2015-07-02
Add apt-vim install recipe

egalpin:
Added recipe for apt-vim installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,37 +42,7 @@ variable, see the documentation for more info.
 ####apt-vim
 Install [apt-vim](https://github.com/egalpin/apt-vim) and Tagbar will be automatically installed as an example. If you already have `apt-vim` but want to add Tagbar, copy and paste the following into your terminal:
 
-```json
-apt-vim install -jy
-{
-    "depends-on": [
-        {
-            "name": "ctags",
-            "recipe": {
-                "darwin": [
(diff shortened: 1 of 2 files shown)


egalpin:
This is more or less a shameless plug for a Vim package management tool that I've been working on, but I wanted to include installation info for it if you'll allow :-)

alerque:
Thanks for the contributions @egalpin. I think including these changes is probably okay, but it makes the README a bit lopsided, as if the only two options for installation are full manual or apt-vim. Granted most users of other plugin managers understand how to use what they use, but I think it would be good to include some other samples here.

Would you be willing to write up a couple more sections and add them to this PR, for example `vim-plug`, `pathogen`, `minpack`, `vundle`, `neobundle`, `dein`, `vam`, etc. It isn't necessary to have them all, but a couple of the popular ones would help balance things out and at the very least mentioning that there are others would be appropriate.

README.md
```diff
@@ -21,8 +21,9 @@ creates the tags it needs on-the-fly in-memory without creating any files.
 [Exuberant ctags 5.5](http://ctags.sourceforge.net/)
 
 ## Installation
-
-Extract the archive or clone the repository into a directory in your
+####Manual
+1. Install [Exuberant ctags 5.5](http://ctags.sourceforge.net/)
```
alerque, line 25: No. Do not install Exuberant, we pretty much only support Universal Ctags now.

PR #306, opened 2016-02-14
Added proper treat of methods in Python.

gryf:
Currently, all functions or methods in tagbar are mapped to function kind.
This patch makes a use of ctags information about the type of the function, so
it will be properly mapped on the structure tree.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -559,11 +559,12 @@ function! s:InitTypes() abort
     let type_python.kind2scope = {
         \ 'c' : 'class',
         \ 'f' : 'function',
-        \ 'm' : 'function'
+        \ 'm' : 'method'
     \ }
     let type_python.scope2kind = {
         \ 'class'    : 'c',
-        \ 'function' : 'f'
+        \ 'function' : 'f',
+        \ 'method'   : 'm'
     \ }
     let s:known_types.python = type_python
     let s:known_types.pyrex  = type_python


timofonic:
No merge?

blueyed:
Hmm.. what version of ctags are you using?

It's not clear to me what these mappings are used for ([doc](https://github.com/majutsushi/tagbar/blob/7b36c46d17d57db34fdb0adac9ba6382d0bb5e66/doc/tagbar.txt#L919-L945)).

I think `m` should be rather mapped to `member` (which is the kind that is being used)?!

Looking at some nested example:

```python
class B():
    def meth_on_B():
        e = 5

        def inner_function():
            class function_in_class:
                inner_member = 1
    b = 2

def foo():
    c = 3

d = 4
```

It looks fine with exuberant ctags (5.8):
```
▼ B : class
  ▼+meth_on_B : function
    ▼+inner_function : function
      ▼ function_in_class : class
          [variables]
          inner_member
    [variables]
    b

 +foo : function

▼ variables
    d
```

Using this patch it turns into:
```
▼ B : class
  ▼ meth_on_B* : function
     +inner_function : function
  ▼+meth_on_B : method
    ▼ inner_function* : function
      ▼ function_in_class : class
          [variables]
          inner_member
    [variables]
    b

 +foo : function

▼ variables
    d
```

Using this patch (but `member` instead):

```diff
@@ -559,11 +561,12 @@ function! s:InitTypes() abort
     let type_python.kind2scope = {
         \ 'c' : 'class',
         \ 'f' : 'function',
-        \ 'm' : 'function'
+        \ 'm' : 'member',
     \ }
     let type_python.scope2kind = {
         \ 'class'    : 'c',
-        \ 'function' : 'f'
+        \ 'function' : 'f',
+        \ 'member'   : 'm',
     \ }
```

With excuberant-ctags (5.8, from Arch Linux) I get:
```
▼ B : class
  ▼ meth_on_B* : function
     +inner_function : function
  ▼+meth_on_B : member
    ▼ inner_function* : function
        function_in_class : class
    [variables]
    b

 +foo : function

▼ variables
    d
```

With universal-ctags (git master), I get:
```
▼+B : class
  ▼+meth_on_B() : member
    ▼-inner_function() : function
       -function_in_class : class
    [variables]
   +b

 +foo() : function

▼ variables
   +d

```

But when using `method` instead of `member`, I get the following with universal-ctags:
```
▼+B : class
  ▼+meth_on_B() : method
    ▼ inner_function* : function
      ▼-function_in_class : class
          [variables]
         +inner_member
    [variables]
   +b

 +foo() : function

 -inner_function() : function

▼ variables
   +d
```

Here `inner_function` is not being mapped into the class.

So the edited patch (using "member") helps with universal-ctags, but breaks exuberant-ctags.
I guess we would have to detect Universal Ctags to address this properly?!

Do you have an example where it is not working correctly for you?

gryf:
@blueyed, I'm using Exuberant Ctags 5.8 both on Gentoo and Ubuntu 14.04 (I can also check on 16.04)
I didn't change it's configuration.

I've took your example, and pass it to ctags:

```shell-session
gryf@combustion ~ $ ctags -f - example.py 
B	example.py	/^class B():$/;"	c
b	example.py	/^    b = 2$/;"	v	class:B
d	example.py	/^d = 4$/;"	v
foo	example.py	/^def foo():$/;"	f
function_in_class	example.py	/^            class function_in_class:$/;"	c	function:B.meth_on_B.inner_function
inner_function	example.py	/^        def inner_function():$/;"	f	function:B.meth_on_B
inner_member	example.py	/^                inner_member = 1$/;"	v	class:B.meth_on_B.inner_function.function_in_class
meth_on_B	example.py	/^    def meth_on_B():$/;"	m	class:B
```

So indeed, B.meth_on_B is indicated as member by ctags, it is not listed twice, and hierarchy is right. Note, that class members are indicated as variables, functions are identified just fine and methods are… members. But that's ok, as long there is differentiate between functions and methods.

As for the tagbar, my intention was to provide separation between function objects and method objects (even though they are identified as 'members' by ctags).  (Exuberant) ctags is doing that just fine, as seen on the shell session above, so the tagbar output is messed up by mechanisms on tagbar itself, that's why changing `method` back to `member` doesn't solve the issue. I'll investigate this further.

majutsushi:
The purpose of the `kind2scope` dictionary is to connect the "kind" information of a containing tag with the extension field of the contained tag that indicates the scope the tag is in. This is somewhat confusing, I assume it's been done this way in ctags for historical reasons.

Here's the Exuberant Ctags 5.9 (svn) output for the example file with all the uninteresting information removed:
```
function_in_class    c    function:B.meth_on_B.inner_function
inner_function       f    function:B.meth_on_B
inner_member         v    class:B.meth_on_B.inner_function.function_in_class
meth_on_B            m    class:B
```

So ctags has assigned the `m` kind to the `meth_on_B()` method, and reports the scope of the `inner_function()` function as `function:B.meth_on_B`. Exuberant Ctags does not report different scopes for a tag depending on whether the containing tag is a method/member or a normal function, it calls both "function" scope.

Here's the relevant part of the Python parser code in Exuberant Ctags:
```
		if (is_class_parent)
		{
			tag.kindName = "member";
			tag.kind = 'm';
			tag.extensionFields.scope [0] = "class";
			tag.extensionFields.scope [1] = vStringValue (parent);
		}
		else
		{
			tag.extensionFields.scope [0] = "function";
			tag.extensionFields.scope [1] = vStringValue (parent);
		}
```

This is why Tagbar maps both the `f` and `m` kinds to the `function` scope, and there's no reverse mapping for `member`, since it can never appear as a scope.

Universal Ctags behaves differently:
```
function_in_class    c    function:B.meth_on_B.inner_function
inner_function       f    member:B.meth_on_B
inner_member         v    class:B.meth_on_B.inner_function.function_in_class
meth_on_B            m    class:B
```

Here it reports the scope as `member` for `B.meth_on_B`. So the proposed patch would actually do the right thing for Universal Ctags, but not for Exuberant Ctags, just like @blueyed said. The term `method` doesn't seem to be used by either ctags implementation.

So the current configuration is correct for Exuberant Ctags, and the proposed patch would be correct for Universal Ctags. This would need a way to check for the exact implementation then.

blueyed:
@majutsushi 
Thanks for your explanations!
Do you think Universal Ctags should be fixed in this regard?
Or for Tagbar to handle a list instead of a One-to-one mapping?

gryf:
@majutsushi Thanks for the explanation. It's pity, that exuberant ctags doesn't provide the right scope information for the members. My initial idea was to distinguish between functions and methods in python. Turns out, it is an issue with exuberant ctags, not tagbar.

majutsushi:
I assume there is going to be an increasing number of differences between Exuberant Ctags and Universal Ctags, so being able to distinguish between them is certainly useful. I've just pushed a change that does that and configures Python correctly based on which version is in use.

PR #308, opened 2016-02-14
Additional function for displaying type of a current tag.

gryf:
Introducing small function for displaying type (kind) of a current tag. Useful
for putting it in the status bar.

This is an alternative version of the #307 - instead expanding the tagbar#currenttag function, new one (tagbar#currenttagtype) was added so the user can influence a way how it will be displayed on the statusbar (or anywhere else).

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -159,6 +159,7 @@ let s:singular_types = {
             \ 'namespaces': 'namespace',
             \ 'net data types': 'net data type',
             \ 'packages': 'package',
+            \ 'package': 'package',
             \ 'paragraphs': 'paragraph',
             \ 'parts': 'part',
             \ 'patterns': 'pattern',
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -122,6 +122,7 @@ let s:singular_types = {
             \ 'domains': 'domain',
             \ 'entities': 'entity',
             \ 'entry points': 'entry point',
+            \ 'embedded': 'embedded',
             \ 'enum constants': 'enum constant',
             \ 'enum types': 'enum type',
             \ 'enumerations': 'enumeration',
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -87,6 +87,122 @@ let s:warnings = {
     \ 'encoding': 0
 \ }
 
+let s:singular_types = {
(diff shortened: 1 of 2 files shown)


gryf:
Probably it's a matter of taste, but I find this particular implementation better than 
#307, it is more comfigurable, and it doesn't change function signature.

autoload/tagbar.vim
```diff
@@ -87,6 +87,124 @@ let s:warnings = {
     \ 'encoding': 0
 \ }
 
+let s:singular_types = {
+            \ 'Classes': 'Class',
```
alerque: Thanks for the contribution @gryf. I just started to look into this and I like the concept of the finished feature but I think the implementation could use a little cleanup before its ready to merge.

For example this array seems like a huge amount of wasted code that doesn't accomplish much. It only changes a couple of types, most of them are pass-through. Would you by chance still be around to review and clean this up some?

gryf: It's just a map from plural form to singular one. This is used for indicating type of the object on the status bar - it would looks awkward if for example there would appear
```
~/somefile.py        MyClass (classes)
```
or
```
~/somefile.py        MyClass.__init__ (methods)
```
instead of `class` and `method`. I'm still around and I'm still actively using this little feature.

alerque: Thanks, and I guess that makes sense. I still think the map is an eyesore, but the real crime here is committed by the English language with its mix of `+s`, `+es`, and even `y→ies` plural transformations. I'll let it slide. :wink:

autoload/tagbar.vim
```diff
@@ -3504,5 +3622,28 @@ function! tagbar#inspect(var) abort
...
+    let s:statusline_in_use = 1
+    let kind = ''
+    let tag = s:GetNearbyTag(0, 1)
+
+    if empty(tag)
+        return a:default
+    endif
+
+    let kind = tag.fields.kind
+    if kind == ''
```
alerque: This should fix the lint error:

```suggestion
    if kind ==# ''
```

gryf: ack.

alerque:
Thanks for the contribution @gryf, and especially for giving it a second (third?) pass so it's all ready to go.

gryf:
No problem. The pleasure is all mine :)

PR #310, opened 2016-02-18
Silent option

nhooyr:
If the silent option is set, tagbar doesn't echo out information about
the tag the cursor is on in the tagbar window.


wsdjeg:
nice ,we need this option .but I think the default var should be 1,just make it works same as before

majutsushi:
Thanks! I've committed a rebased version due to a merge conflict.

PR #311, opened 2016-02-24
Verify that window number is valid before switching to it

### LINKED ISSUE #305: Error when quit file while tagbar open
I got the following error when I try to close file while tagbar is still open:

screen shot 2016-02-04 at 7 46 38 pm

I am using tagbar with macvim.

joegrasse:
Fixes #305

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3941,8 +3941,15 @@ function! s:QuitIfOnlyWindow() abort
         endif
     endif
 
-    call s:goto_win(prevwinnr, 1)
-    call s:goto_win(curwinnr, 1)
+    " Check that prevwinnr is still a valid window number
+    if prevwinnr <= winnr('$')
+      call s:goto_win(prevwinnr, 1)
+    endif
+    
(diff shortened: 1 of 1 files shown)


codinfox:
@joegrasse Unfortunately this leads to error on macvim:

```
Error detected while processing /Users/Codinfox/.vim/bundle/tagbar/plugin/tagbar.vim:
line    1:
E169: Command too recursive
Tagbar: Could not load plugin code, check your runtimepath!
```

joegrasse:
Interesting. Not sure what would cause the recursive loop to happen. Is there any other differences between your tagbar.vim and the one in the pull request?

codinfox:
I directly used your pull request. :(

dperny:
This PR fixes a bug related to NERDTree. If NERDTree is opened, the main text editing split switched to, Tagbar is opened, and then the main editing split is closed without switching between any splits, an error is thrown.

majutsushi:
Thanks! I've merged your pull request after a small change due to a conflict.

Issue #331, opened 2016-08-26
Case-independent sort


jpmv27:
I am working in "C" locale, if it matters
Names are sorted with upper-case letters first and lower-case letters last
Is there any way to configure a case-independent sort?

majutsushi:
I guess this can be closed now that #333 is merged.

PR #333, opened 2016-08-27
Add case-insensitive sort option

jpmv27:
Implements the option I was looking for in #331


majutsushi:
Thank you for the very comprehensive pull request, especially the documentation everywhere! There's nothing I could have improved there :)

Issue #332, opened 2016-08-27
hide_nonpublic option doesn't seem to do anything


jpmv27:
Perhaps I am misunderstanding what this option is supposed to do
With this option set to 1 in .vimrc, functions declared static are still visible in tagbar of C file.
Using 'v' to toggle the setting has no visible effect, except showing the [v] flag in the status bar

jpmv27:
It looks like the option only affects scoped types, not non-scoped types. Is that intentional?

majutsushi:
This option relies on the `access` extension field from ctags to determine whether a tag is public or not. Unfortunately ctags doesn't add that `access` field to `static` tags in C, so Tagbar has no way of knowing that these tags should be considered non-public, and it wouldn't even be possible (in a reliable way) for Tagbar to parse the tag information from ctags in a way that would let it figure that out by itself. So unfortunately the only way to fix this would be to change ctags.

PR #334, opened 2016-08-27
Keep cursor on same tag when toggle hide non-public

jpmv27:
When toggling the 'v' option, cursor was not staying on the same tag after the list has been updated
Ran into this while looking further into behaviour of 'v' (re. #332)


majutsushi:
Thank you!

Issue #330, opened 2016-08-26
Incorrect file name in status bar in tagbar window


jpmv27:
When I first toggle tagbar on, the file name in the status bar of the tagbar window matches the name of the file from which I opened tagbar

If I then open a new file in the main window, the tagbar information is updated but the file name in the status bar isn't. This can be confusing, especially if TagbarTogglePause

jpmv27:
Same issue happens when using :TagbarTogglePause.
If I pause, change file, then un-pause, the tagbar contents are updated to match the new file but the status line still shows the previous file

Should AutoUpdate() call SetStatusLine()? Or should callers of AutoUpdate() also call SetStatusLine() when appropriate?

majutsushi:
This should be fixed now, thanks for the report!

jpmv27:
I'm not sure the fix addresses the whole of the problem. I was also seeing the same issue when tagbar is open and I start editing a new file: the tagbar contents are updated but the filename in the status bar is not. Is that fixed too?

MV
P.S. My own proposed fix is in PR#335

majutsushi:
Hmm, that seems to be working fine for me. Is it not working for you?

I did see your pull request, thanks for that! And also for noticing that the argument was redundant. As far as I can see both of the changes related to this issue are explicitly updating the statusline after a call to `AutoUpdate`, so calling it at the end of every `AutoUpdate` should cover both of those cases. Your changes would certainly be more efficient than calling it after every update, however the statusline can contain more information than just the file name and may thus need to update more often, and it shouldn't be an expensive call anyway. If it becomes a performance issue it could be changed later. Is that okay with you?

jpmv27:
Never mind, I misread your commit.

Thanks
MV

PR #335, opened 2016-08-28
Update status line when changing buffer or unpausing

jpmv27:
Proposed fix for #330 
The "current" parameter to SetStatusLine was redundant so I removed it, which made the new code simpler


majutsushi:
The issue should now be fixed in #330, and I've used your removal of the redundant argument, so I guess this can be closed.

PR #338, opened 2016-09-10
Add asymptote support as a c clone

alejandrogallo:
Asymptote support which is basically a stable construct copied verbatim
from the type_c variable. It parses all basic variable types well,
and it could be extended in the future.


majutsushi:
Thanks!

alejandrogallo:
Thank you for the great work!

PR #342, opened 2016-09-24
Separate tagbar per tab

jpmv27:
Hi,

Not sure whether you'll be interested in this. In my workflow, it's handy to be able to have a different tagbar on different tabs, so I created these commits.

MV

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1574,7 +1574,7 @@ endfunction
 
 " s:KindheaderTag.toggleFold() {{{3
 function! s:KindheaderTag.toggleFold() abort dict
-    let fileinfo = s:state.getCurrent(0)
+    let fileinfo = s:TagbarState().getCurrent(0)
 
     let fileinfo.kindfolds[self.short] = !fileinfo.kindfolds[self.short]
 endfunction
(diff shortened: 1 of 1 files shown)


majutsushi:
I'm definitely interested in supporting multiple tabs, I work that way too occasionally. I has worked fine for me in the past though, so I'm curious what issue you were having that require the changes you made. Could you give a quick example of wasn't working for you before? I'm just trying to understand the problem better. Thanks!

jpmv27:
Very interesting.

I recall I was running into some issues (I forget the details now,) but
trying it again briefly now with 'stock' tagbar I do not see any issues.

Perhaps they were resolved by some other fixes.

Let's keep the PR on hold while I look into it more carefully.

Regards,
MV

jpmv27:
Well, that's embarassing :-)

Either there was some other plug-in interfering, or I was chasing ghosts, because I don't see any issue with using tagbar simultaneously in multiple tabs.

I'll close this PR now...
MV

jpmv27:
OK< I see the problem now. Not sure it justifies these changes, I will open an issue and we can discuss there

jpmv27:
New commit solves #345 by leveraging tab-scope variable introduced to fix #344

majutsushi:
Thanks for the explanations, and for all your work! I've confirmed the issue you described in #344 and that the pull request fixes it. I guess #344 can be closed as well then?

jpmv27:
Yes, thanks.
MV

Issue #322, opened 2016-06-24
the tagbar is closing the gvim by the commad ":bd"


bigov:
Hi! I'm use gvim for C++ development. I found one problem: when deleting one of the buffers by command ":bd", the opened tagbar is closing gvim session with the main window and all other multiple early opened buffers. If you need, in the archive my complete vim's config. My Vim version is 7.4 (1829).

[vimfiles.zip](https://github.com/majutsushi/tagbar/files/331344/vimfiles.zip)

majutsushi:
As mentioned in #234 this should finally be fixed now.

Issue #291, opened 2015-09-19
cannot toggle Tagbar when using netrw


gavinchou:
when using netrw to browse a direcotry:

    vim /some/path/of/directory

or, in vim

    :e /some/path/of/directory

and then open a file and toggle tagbar with

    :Tagbar

then, tagbar wont work, tagbar flashes and quits
the problem occurs in the following function (autoload/tagbar.vim), 
` call s:goto_win(tagbarwinnr, 1)`

    " s:QuitIfOnlyWindow() {{{2
    function! s:QuitIfOnlyWindow() abort
        let tagbarwinnr = bufwinnr('__Tagbar__')
        if tagbarwinnr == -1
            return
        endif
    
        let curwinnr = winnr()
        let prevwinnr = winnr('#') == 0 ? curwinnr : winnr('#')
        call s:goto_win(tagbarwinnr, 1)

i (temporally) fix it by modifying this function,

    " s:QuitIfOnlyWindow() {{{2
    function! s:QuitIfOnlyWindow() abort
       if bufwinnr('__Tagbar__') == winnr() && tabpagewinnr(tabpagenr(), '$') == 1
            exe "q"
        endif
        return
    endfunc

 it seems i fixed, maybe there is a better way?

ps. it seems the logic of the original script is fine, but somehow it is not, lets fix it!

symbolix:
I am also experiencing issues related to loading files through the `netrw`. Once I select the file to be loaded or switched to, I get the following error:

    Error detected while processing function <SNR>102_QuitIfOnlyWindow[28]..<SNR>102_goto_win:
    line    8:
    E16: Invalid range: 2wincmd w

Is it possible to fix this? I am using VIM 7.4.1553 with `netrw` v155. `netrw` is a core component of VIM. This plugin should be compatible with VIM's native file browser.

qwIvan:
+1

ggaaooppeenngg:
+1

SoPudge:
i have the same problem

majutsushi:
I think this should be fixed now. Sorry for the long delay.
If there's still a problem please reopen this issue.

PR #343, opened 2016-09-25
Rework quitifonlywindow

jpmv27:
Meant to send this PR a while ago, but it slipped my mind

While working to support tagbar-per-tab, I noticed an issue when with doing ":tabe" when the tagbar is open. Has similar behaviour as described in #291 for netrw. After fixing that, I decided to see whether I could address #322. The solution is not perfect, but at least vim no longer exits unexpectedly.


jpmv27:
Will review this PR as well and resubmit if necessary

PR #349, opened 2016-10-11
flags argument added to functions responsible for window toggling

vavaka:
It would be great to have the ability to pass custom flags to OpenWindow function from ToggleWindow function. In this case users will be able to customize Tagbar behaviour in "toggle" mode. For example in my configuration I want:
* Tagbar be toggled by pressing the same key (<C-2> in my case);
* cursor be moved to Tagbar window when Tagbar window becomes visible;
* Tagbar window automatically be closed on tag selection.

Having changes from this pull request I can setup Tagbar behaviour I want:
```
command! -nargs=0 TagbarToggleFocusAutoClose call tagbar#ToggleWindow('fcj')
nmap <C-2> :TagbarToggleFocusAutoClose<CR>
```

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1775,7 +1775,7 @@ endfunction
 
 " Window management {{{1
 " s:ToggleWindow() {{{2
-function! s:ToggleWindow() abort
+function! s:ToggleWindow(flags) abort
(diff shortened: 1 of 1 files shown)


majutsushi:
That's quite useful, thanks!

PR #360, opened 2016-11-03
Escape filename in ShowInPreviewWin()

ChesleyTan:
File paths containing special characters or spaces are not escaped causing `pedit` to fail.


majutsushi:
Thanks!

PR #368, opened 2016-11-26
Add possibility of custom vimscript function to rewrite/filter tags.

gasparch:
This patch allows to have vimscript function which can rewrite tag names.

I use it for Elixir plugin where I remove duplicate function heads (it is annoying to have 10 'process' functions while it is just different clauses of same function). Also it allows me to add function arity to tags, which immediately make them more useful.

I think this case is very specific to Elixir/Erlang programming, but may be someone else will also benefit from using that functionality.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3236,6 +3236,70 @@ function! s:JumpToTag(stay_in_tagbar) abort
     endif
 endfunction
 
+" s:JumpToTagName() {{{2
+function! s:JumpToTagName(tag_name) abort
+    let taginfo = s:GetTagInfoByName(a:tag_name)
+
+    if empty(taginfo) || !taginfo.isNormalTag()
+        return
+    endif
+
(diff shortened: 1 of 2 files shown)


majutsushi:
That's a very interesting idea, something like that could be quite useful. Before I include something like this that manipulates a currently internal data structures I would like to make sure there is a reasonably future-proof API with all the relevant documentation. Especially since I recently had an idea about how to change the parsing in a way to make it hopefully both faster and simpler, which would result in that particular list going away. But I haven't nailed down all the details yet.

So I'm not going to merge this as of yet, but I'll definitely keep it in mind looking at the parsing changes and will try to make something like this possible.

gasparch:
okay, then I will keep fork for some time now until it is possible to do it in less intrusive way.
another possibility for me will be writing custom wrapper around crags or elixir xref/code exploring server and feeding correct data into tagbar.

PR #371, opened 2016-12-01
Adds FocusBuffer so you can jump to the tagbar buffer from any window.

kyleterry:
This is very useful for vertical splits. I sometimes want to explore the tags from a window on the left
of a two or more vertical split and navigating to the tagbar buffer with the vim defaults causes the tagbar
context to change. This resolves that by adding a function that will let you focus the buffer from anywhere.


majutsushi:
This should already be possible with `:TagbarOpen j`. Is that not working for you?

kyleterry:
Somehow I missed the 'j' in the docs. Sorry for wasting your time.

Thank you.

PR #372, opened 2016-12-02
  Clojure support improvement

micrub:
Added kinds:

  * namespace
  * definition
  * function
  * private function
  * macro
  * inline
  * multimethod definition
  * multimethod instance
  * definition (once)
  * struct
  * intern
  * namespace


majutsushi:
Thanks! I'm currently planning to separate the Exuberant Ctags and Universal Ctags support to more easily support both without compatibility issues. I'll include this in the process.

Issue #383, opened 2016-12-18
HandleBufDelete aborts with errors


jpmv27:
When using tagbar along with other vim plugins, some operations can cause `HandleBufDelete` to fail. The plugins are BufExplorer and vim-project

Relevant (maybe) configuration:
- set hidden
- set switchbuf=usetab

SCENARIO 1 (vim-project)
- Open a C file in a buffer: `:e file.c`
- Toggle tagbar on
- In window of C file buffer, issue `:Welcome` command
- Result: `E937: Attempt to delete a buffer that is in use`

SCENARIO 2 (BufExplorer)
- Open a C file in a buffer: `:e file1.c`
- Open a second C file in a second buffer in the same window: `:e file2.c`
- Toggle tagbar on
- In window of C file buffer, issue `:BufExplorer` command
- Result: `E93: More than one match for [BufExplorer]`

In both cases, `HandleBufDelete` is called by a `BufDelete` autocmd triggered as a result of the plugin executing `set nobuflisted`.

Please refer to PR #384. This work-around avoids these errors, and doesn't seem to cause any other issues, but I'm not sure whether this is a legitimate fix.

majutsushi:
I guess this can be closed now as well.

PR #384, opened 2016-12-18
HandleBufDelete: Ignore apparent 'set nobuflisted' events

jpmv27:
Please refer to #383


majutsushi:
Thanks for this! This seems to be the best solution to the problem.
Just one small niggle: Instead of using a `+` prefix to coerce the string into a number it would be better to use the `str2nr()` function, which would also make the comment unnecessary.

jpmv27:
On 12/26/16 20:57, Jan Larres wrote:
> Thanks for this! This seems to be the best solution to the problem.
> Just one small niggle: Instead of using a `+` prefix to coerce the string into a number it would be better to use the `str2nr()` function, which would also make the comment unnecessary.
>

Quite right. I have updated the PR
MV

majutsushi:
Thanks!

PR #394, opened 2017-02-06
Add note about Universal ctags

fievel:
I found that exhuberant tag seem abandonned for long time (last version is from 2009), however a fork as been made called Universal ctags, it seems to work great, correct few bugs and add additional langages. I think it could be useful to advice other users to use it.


majutsushi:
Thanks, and sorry for the long wait! I wanted to make sure that the changed definitions in Universal Ctags were properly supported first.

PR #406, opened 2017-03-14
Fix scopes for ruby

ToruIwashita:
Currently scope is always class. This change fixes it

### before
image

### after
image


majutsushi:
Thanks, but unfortunately this won't work. The values in that dictionary are not simply display names, they are references to the scope fields that ctags uses to report the scope that a tag is in, so changing their values will result in tags appearing in a wrong scope. As mentioned in #407, I have some longer-term changes planned that will fix this properly, I just can't promise a timeline yet.

ToruIwashita:
I understood. Thank you for your review!

PR #423, opened 2017-06-27
compatibility with phpctags

jolan78:
https://github.com/techlivezheng/phpctags

phpctags --version output :

Version: 0.5.1

Exuberant Ctags compatiable PHP enhancement, Copyright (C) 2012 Techlive Zheng
Addresses: <techlivezheng@gmail.com>, https://github.com/techlivezheng/phpctags


alerque:
Thanks for the contribution @jolan78! I'm sorry it took so long but the change looks good to me.

PR #426, opened 2017-07-26
Add ability to open or close folds a single level

AdnoC:
I often want to see the second-level tags or all but the last fold of tags. This PR adds the ability to do that.

The new keybinds `zr` and `zm`  act like they do with normal folds. It increases or decreases the buffer's foldlevel by 1 and opens/closes folds to match it.

In order to make `zm` it first checks whether the fold level might be something set by the user (i.e. not `99`). If it is the max value (`99`), it lowers the fold level to as low as possible while still having all folds open. An alternative could be to check whether decreasing the fold level resulted in a fold closing, and repeatedly reducing the level until at least one fold closed. I felt that just checking for a foldlevel of `99` would suffice though.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1031,6 +1031,8 @@ function! s:MapKeys() abort
         \ ['togglefold',    'ToggleFold()'],
         \ ['openallfolds',  'SetFoldLevel(99, 1)'],
         \ ['closeallfolds', 'SetFoldLevel(0, 1)'],
+        \ ['incrementfolds',  'ChangeFoldLevel(1, 1)'],
+        \ ['decrementfolds',  'ChangeFoldLevel(-1, 1)'],
         \ ['nextfold',      'GotoNextFold()'],
         \ ['prevfold',      'GotoPrevFold()'],
         \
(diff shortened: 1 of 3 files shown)


majutsushi:
Thanks, this looks very good! You even thought about the documentation :)

Just one thing that I think would make it even nicer: Currently the `ChangeFoldLevel()` function is mostly a copy of the `SetFoldLevel()` one. I think it would make sense if the `change` function just calculated the new level and then passed it on to the `set` function, to reduce code duplication a bit.

I noticed that the `MinimizeMaxFoldLevel()` function is a bit unusual in how it extends the collection with nested tags. I would have used a recursive function with an accumulator instead, but I guess this works as well.

AdnoC:
You're right. Not sure why I copy-pasted the entirety of `SetFoldLevel()` when making `ChangeFoldLevel()`. Changing it like you suggested

AdnoC:
As for extending the collection instead of using recursion in `MinimizeMaxFoldLevel()`... I don't know. I guess I just prefer an iterative solution for this. I don't do much VimL outside of my vimrc though, so if its strange or you don't like it I can redo it based of the `SetFoldLevelRecursive()` for a recursive solution.

majutsushi:
Thanks for the update! And don't worry about the other function, it's just something I noticed.

PR #430, opened 2017-08-18
Add PHP scopes and additional kinds

weirdan:
Added scopes, so tagbar will show the following hierarchy:
```
  namespace
    subnamespace
      ...
        class
	  property (variable)
	  ...
	  method
	  ...
```

screenshot fri aug 18 2017 03 29 46 gmt 0300 eest 1913x971

Additional kinds added (basically the same as #283):
 * namespaces
 * use aliases (imports)
 * traits


majutsushi:
Thank you! I had to merge it locally due to some conflicts after a bit of refactoring. I removed the `j` kind since it doesn't seem to be used any more.

I also made some changes so that the namespaces don't get displayed separately any more, but instead have the proper hierarchy. Which was trickier than expected.

PR #435, opened 2017-08-29
support NuWeb

dangkhoasdc:
Support [nuweb](http://nuweb.sourceforge.net). I simply force Tagbar to load Latex filetype. 

P.s: to properly run with ctags, append the following line to `~/.ctags`:

```
--langmap=Tex:+.nw
```


majutsushi:
Thanks! I'm currently thinking about how to specify types that need additional ctags configuration in a way that makes things work out of the box, without having to add these kinds of definitions manually. So I'm going to hold off with merging the pull request until this is in place.

PR #444, opened 2017-09-18
fixed encoding bug of balloon

faster-mo:
Balloon show garbled text in gvim when tag prototype include cjk encoded character and system_encoding not equal to &encoding.
image


majutsushi:
Thanks!

PR #447, opened 2017-09-23
Ease debugging

teto:
These are changes I made to help me debugging my universal-ctags problem but I think they can be helpful in generla. 

When debug is enabled
- add verbose flag to ctags
- don't remove temp files
- log the temporary file created

- also check for writefile returned value


majutsushi:
Thank you!

PR #458, opened 2017-11-20
Rust: fix method kind from F to P for universal-ctags

ithinuel:
Universal ctags uses the short kind F for a special meaning and uses P for rust's method.


Luxed:
Thank you for this PR, I was searching what was wrong as the output of ctags seemed to be correct.
Everything works with this.

majutsushi:
Thanks!

PR #470, opened 2018-01-11
Add tagbar#StopAutoUpdate

blueyed:
This is useful to stop updating information, after using
`tagbar#currenttag` manually/temporarily only.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -862,8 +862,7 @@ function! s:CloseWindow() abort
     call s:ShrinkIfExpanded()
 
     if s:autocommands_done && !s:statusline_in_use
-        autocmd! TagbarAutoCmds
-        let s:autocommands_done = 0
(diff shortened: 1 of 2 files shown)


alerque:
Thanks for the contribution @blueyed! It took me a bit to figure out what it was doing, but this does look good.

wsdjeg:
@alerque maybe we should use rebase merge or squash merge to keep the git commit history not to be changed.

alerque:
@wsdjeg In some cases yes that is helpful, but that also makes it very difficult to tell from Github's fork browsing tools what things have been integrated and what hasn't. There are a lot of assorted fixes hanging out there that never ever got PRed (some of this can be attributed to the PRs that were here being clearly unattended) and I would like to look through them eventually and see if there are any gems we're missing out on.

wsdjeg:
If the PR branch is not master of the forked repo, it can be squashed and merged.

PR #471, opened 2018-01-17
Fix issue where --regex based file types can output duplicate lines when --sort=no

themadsens:


alerque:
Thanks for the contribution @themadsens! This looks good to me. A little rare perhaps in practice but worth dodging. Note I'm squashing this PR because the second commit just fixes something the first one introduces and should probably just have been `--ammend`ed to it anyway. Just a note for next time ;-)

PR #481, opened 2018-02-22
Use function! to allow for reloading

blueyed:
Fixes the following when reloading the file:

> E122: Function <SNR>188_ProcessTag already exists, add ! to replace it

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1255,7 +1255,7 @@ function! s:ParseTagline(part1, part2, typeinfo, fileinfo) abort
 endfunction
 
 " s:ProcessTag() {{{2
-function s:ProcessTag(name, filename, pattern, fields, is_split, typeinfo, fileinfo) abort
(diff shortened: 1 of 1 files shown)


alerque:
Thanks again @blueyed. This actually got fixed in a little more depth in #471 which I just merged, but Github still thinks this can be merged. I'm not actually sure why, but I suspect it will figure out there is nothing to do when I try to merge. I expect it to fail because it would be an empty change set, in which case we'll just close this. In the event I'm the one with things cross-wired, lets see what happens...

alerque:
...oops. I think that should be a bug in GitHub actually, in should be able to see when a merge is a no-op. If only they had an issue tracker.

blueyed:
Thanks!
You could have just closed it also as far as I'm concerned.

alerque:
@blueyed I realize that, I actually wanted to see what would happen and it's hard (not impossible, just time consuming) to drum up artificial examples.

blueyed:
@alerque I see, thanks for letting me be part of this experiment.. :)

PR #515, opened 2018-12-02
Add TagbarJump

ggVGc:


plugin/tagbar.vim
```diff
@@ -146,6 +146,8 @@ command! -nargs=1 TagbarGetTypeConfig call tagbar#gettypeconfig(<f-args>)
 command! -nargs=? TagbarDebug         call tagbar#debug#start_debug(<f-args>)
 command! -nargs=0 TagbarDebugEnd      call tagbar#debug#stop_debug()
 command! -nargs=0 TagbarTogglePause   call tagbar#toggle_pause()
+command! -nargs=0 TagbarJump   call tagbar#jump()
```
alerque, line 149: Somehow (I'm not quite sure how) this command needs to only be available in Tagbar windows. In other places it throws an error.

raven42: Can we add the following into the `tagbar#jump()` routine to address this concern?
```
if $filetype !=# 'tagbar'
    " Not in tagbar window - ignore this function call
    return
endif
```

raven42 (approved):

ggVGc:
Hi, sorry for not being active on this. Other things have been getting in the way everytime I got reminded by a notification to take a look, and then I forgot about it again every time.
Glad it's getting accepted, even if late! (Sorry for not adding the docs)

alerque (approved):

PR #529, opened 2019-03-26
Add asciidoc support

hupfdule:
universal ctags supports asciidoc (apart from [a bug that may leave out the first level heading](https://github.com/universal-ctags/ctags/issues/2061)).

This PR adds asciidoc support to tagbar.

However it is not perfect. The output is not what I would expect it to be. I compare it with the support for latex which is _very_ similar.

adoc_vs_tex

adoc
tex

As you can see the items for the latex file are correctly structured in a full tree structure.

But the items for the asciidoc file only contain the direct children. And those children are then displayed again at the top level.

Also the anchors in the asciidoc file are displayed only in the sections they are defined in whereas they are in a separate top level structure "labels" for the latex file.

I actually don't expect this to be the result of the aforementioned bug, since then I would expect only a problem with the level 1 headings (that are not visible, as expected).

Any help to fixing this is appreciated!

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -47,6 +47,37 @@ function! tagbar#types#uctags#init(supported_types) abort
         \ {'short' : 't', 'long' : 'targets',    'fold' : 0, 'stl' : 1}
     \ ]
     let types.ant = type_ant
+    " Asciidoc {{{1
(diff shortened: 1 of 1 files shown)


hupfdule:
It seems that the above mentioned problem is another bug in the ctags asciidoc parser. See https://github.com/universal-ctags/ctags/issues/2063.

PR #548, opened 2019-10-10
Add g:tagbar_visibility_symbols

villainy:
Allow for custom visibility symbols.

### DIFF
--- a/autoload/tagbar/prototypes/basetag.vim
+++ b/autoload/tagbar/prototypes/basetag.vim
@@ -1,9 +1,3 @@
-let s:visibility_symbols = {
-    \ 'public'    : '+',
-    \ 'protected' : '#',
-    \ 'private'   : '-'
-\ }
-
 function! tagbar#prototypes#basetag#new(name) abort
     let newobj = {}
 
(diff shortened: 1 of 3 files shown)


autoload/tagbar/prototypes/basetag.vim
```diff
@@ -85,9 +79,9 @@ function! s:_getPrefix() abort dict
     " Visibility is called 'access' in the ctags output
     if g:tagbar_show_visibility
         if has_key(self.fields, 'access')
-            let prefix .= get(s:visibility_symbols, self.fields.access, ' ')
+            let prefix .= get(g:tagbar_visibility_symbols, self.fields.access, ' ')
```
alerque: This doesn't look like it will work quite right to me. Setting custom ones as in your use case will work because `g:tagbar_visibility_symbols` will be defined, but in the case it is not, the default values entered here will nod be found because they are in in a different array. This needs to fallback to the default if an override is not found.

alerque:
Thanks again for taking the time to contribute. Unfortunately since you didn't get around to fixing the outstanding issues, somebody else came in and made this obsolete with another PR. The bad part of that is some work was duplicated / wasted. The good news is that this feature is now available to everybody. Enjoy!

PR #561, opened 2019-10-30
A simple way to implement tagbar async call

pangchol:


autoload/tagbar.vim
```diff
@@ -354,6 +354,7 @@ function! s:CreateAutocommands() abort
         if exists('##QuitPre')
             autocmd QuitPre * let s:vim_quitting = 1
         endif
+        " ������Զ��������ڸ��˻��������ò���, �����׵��´��ļ�ʱ��Ļ�������ˢ��,���Թر�
```
alerque: Please add comments in English (even if it's Google Translate) so the rest of us can figure out what this is about.

autoload/tagbar.vim
```diff
@@ -2413,12 +2414,24 @@ endfunction
...
 " Helper functions {{{1
 " s:AutoUpdate() {{{2
+" use timer_start to let tagbar async, this can increase vim open file performance and
+" fix windows blink when open some file.
 function! s:AutoUpdate(fname, force, ...) abort
-    call tagbar#debug#log('AutoUpdate called [' . a:fname . ']')
+    let g:tagbar_update_fname = a:fname
+    let g:tagbar_update_force = a:force
+    let g:tagbar_no_display = a:0 > 0 ? a:1 : 0
+    if has('win32') " windows use timer_start will call bug
```
alerque: Where is this bug documented upstream (so we can track when to revert workarounds)

autoload/tagbar.vim
```diff
@@ -2413,12 +2414,24 @@ endfunction
...
+        call timer_start(0, 'AutoUpdate_CB')
+    endif
+endfunc
 
-    " Whether we want to skip actually displaying the tags in Tagbar and only
-    " update the fileinfo
-    let no_display = a:0 > 0 ? a:1 : 0
+function! AutoUpdate_CB(channel, ...) abort
+    " call tagbar#debug#log('AutoUpdate called [' . fname . ']')
+    let l:fname = g:tagbar_update_fname
```
alerque: Why are you passing global variables around here? This seems to rely on side effects, this function should be refactored to accept all the arguments it will use (as in the function it is replacing).

plugin/tagbar.vim
```diff
@@ -95,31 +95,31 @@ endif
...
-        \ ['openfold',      ['+', '<kPlus>', 'zo']],
-        \ ['closefold',     ['-', '<kMinus>', 'zc']],
-        \ ['togglefold',    ['o', 'za']],
-        \ ['openallfolds',  ['*', '<kMultiply>', 'zR']],
-        \ ['closeallfolds', ['=', 'zM']],
-        \ ['incrementfolds',  ['zr']],
-        \ ['decrementfolds',  ['zm']],
-        \ ['nextfold',      'zj'],
-        \ ['prevfold',      'zk'],
+        \ ['openfold',       '+'],
```
alerque: Why did you drop all the standard keymaps (that match vim's defaults?)

pangchol: Sorry, the second commit was originally only for my local use. As I was not familiar with the function of pull requst, the second commit error was more closely related to pull request. I originally just wanted to submit the first pull request.

autoload/tagbar.vim
pangchol: The use of global variables is because I don't know how to pass parameters in a callback , so I'm just using global variables for simple pass paramenters.

alerque: We're going to have to figure out how to do that, or at least some other mechanism which will protect them from being clobbered. As soon as asynchronous jobs start getting involved it's really easy to introduce bugs if you are relying on side effects. I don't want to start down that road here.

plugin/tagbar.vim
alerque: Okay. You can remove it by resetting your branch to the correct commit and force-pushing. Keep in mind Github will keep this PR up to date with your branch, so you should do the work for the PR in a different branch than the one you are using. This is why Github shows you a big error message when you open PRs against your master branch saying that is an anti-pattern!


My suggestion would be to start a second branch for your own use, then tweak this one to remove that commit (and also address other issues brought up in review here.

autoload/tagbar.vim
alerque: It looks like you can pass a function reference, so something like:

```nvim
call timer_start(1, function('AutoUpdato_CB', [a:jobid], fname))
```

But you'll have to play around with specifics. The [async.vim](https://github.com/prabirshrestha/async.vim) code might give you some clues.

raven42: so after some experimentation, the `timer_start()` routine operates in the same thread context. So this would only delay the processing of ctags, and any apparent 'lockup' effect of vim during ctags processing will still occur, only a little bit after once the timer pops.

PR #571, opened 2019-12-22
Fall back to using bourne shell if $SHELL is Elvish

vendion:
If vim is launched from the Elvish shell, then fall back to using the
bourne shell (sh) or a bourne compatible shell when calling
{universal/exuberant-}ctags.

Signed-off-by: Adam Jimerson <vendion@gmail.com>


alerque:
Thanks for taking the time to contribute this.

PR #575, opened 2020-01-02
Test whether &termencoding even exists before use

### LINKED ISSUE #574: g:tagbar_iconchars handling breaks on latest Neovim
The "termencoding" option has been removed from the latest neovim (0.5.0 master branch). Tagbar expects this variable to exist if `g:tagbar_iconchars` is not set in the user's initialization file. See below for the error message:

unknown_option_termencoding

Temporary fix:

```viml
" init.vim
let g:tagbar_iconchars = ['▶', '▼']
```

alerque:
Fixes #574

See also:

* https://github.com/neovim/neovim/pull/2631
* https://github.com/neovim/neovim/issues/7445
* https://github.com/vim/vim/commit/ac360bf2ca293735fc7c6654dc2b3066f4c62488

### DIFF
--- a/plugin/tagbar.vim
+++ b/plugin/tagbar.vim
@@ -86,7 +86,7 @@ call s:setup_options()
 
 if !exists('g:tagbar_iconchars')
     if has('multi_byte') && has('unix') && &encoding ==# 'utf-8' &&
-     \ (empty(&termencoding) || &termencoding ==# 'utf-8')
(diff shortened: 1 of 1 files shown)


alerque:
I don't have Windows to test this on, can somebody confirm that we can send Unicode stuff there in the event the `&termencoding` option does not exist?

jcorbin:
Confirmed that this Works For Me ™ in neovim

PR #576, opened 2020-01-02
Fix termencoding check

jcorbin:
Fixes usage in neovim, as this is a dropped option.

Also maybe on Windows, since `vim_diff.txt` in neovim says "Vim 7.4.852 also removed this for Windows".

### DIFF
--- a/plugin/tagbar.vim
+++ b/plugin/tagbar.vim
@@ -86,7 +86,7 @@ call s:setup_options()
 
 if !exists('g:tagbar_iconchars')
     if has('multi_byte') && has('unix') && &encoding ==# 'utf-8' &&
-     \ (empty(&termencoding) || &termencoding ==# 'utf-8')
+     \ (!has('termencoding') || empty(&termencoding) || &termencoding ==# 'utf-8')
         let g:tagbar_iconchars = ['▶', '▼']
     else
         let g:tagbar_iconchars = ['+', '-']


alerque:
Thanks for taking the time to contribute ... but did you by chance see #575 before working on this?

plugin/tagbar.vim
```diff
@@ -86,7 +86,7 @@ call s:setup_options()
 
 if !exists('g:tagbar_iconchars')
     if has('multi_byte') && has('unix') && &encoding ==# 'utf-8' &&
-     \ (empty(&termencoding) || &termencoding ==# 'utf-8')
+     \ (!has('termencoding') || empty(&termencoding) || &termencoding ==# 'utf-8')
```
alerque: I tested this on by Neovim bus it does not identify when the option does exist. This is not a feature flag (what `has()` usually tests for), it is an option. According to the docs I found `exists()` is the way to check this, with either `&...` for just whether the option exists or `+...` for whether it's actually a functional option (as opposed to legacy deprecated or whatever). Can you check the other PR and make sure it works for you?

jcorbin:
Ahh great point, no I looked  at issues briefly, should have checked open PRs.

alerque:
I can confirm that this does _not_ work for me on a version of vim that does have this option. It suppresses the error for you because it suppresses it for _everybody_ because `has()` will never return true here.

As soon as anybody can confirm that the other PR doesn't error out on the other end (where the option has been dropped) I'll merge it. I can only test the other end of things at the moment.

PR #579, opened 2020-01-03
Add g:tagbar_show_balloon option

zhmars:
This adds an option `g:tagbar_show_balloon`, which allows to disable balloon messages.


alerque:
Thanks for this contribution @zhmars! It looks great to me, and special thanks for adding corresponding documentation too.

zhmars:
Thanks for your quickly reply and maintaining.

PR #581, opened 2020-01-03
feat(dart_ctags):  support for dart if dart_ctags is installed

### LINKED ISSUE #570: Support for Dart
I'm trying to add Dart to Tagbar.

I'm using [dart_ctags](https://github.com/nerdrew/dart-ctags) for the tags, and they generate these functional [tags](https://pastebin.com/7h1TW2gR).

I've added to the Tagbar config:
```
let g:tagbar_type_dart = {
	\ 'ctagstype' : 'dart',
	\ 'kinds'     : [
		\ 'c:classes',
		\ 'i:types',
		\ 'm:methods',
		\ 'M:constructor',
		\ 'f:functions',
	\ ],
	\ 'ctagsbin'  : 'pub',
    \ 'ctagsargs' : 'global run dart_ctags:tags',
\ }
```

Tagbar is able to run this successfuly, however, I get these error messages in the vim console:

```
Error detected while processing function <SNR>165_AutoUpdate[63]..<SNR>165_ProcessFile[121]..<SNR>169_sortTags[2]..tagbar#sorting#sort[6]..<SNR>173_compare_by_kind:
line    3:
E716: Key not present in Dictionary: c^I
E15: Invalid expression: typeinfo.kinddict[a:tag1.fields.kind] <# typeinfo.kinddict[a:tag2.fields.kind]
E716: Key not present in Dictionary: f^I
E15: Invalid expression: typeinfo.kinddict[a:tag1.fields.kind] <# typeinfo.kinddict[a:tag2.fields.kind]
```

Here is the [tagbar debug log](https://pastebin.com/28dY0xc8).

Does anyone have any ideas about handling the f^I or c^I missing key errors?

jrc2139:
still waiting on the updated code for dart_ctags to get published on the dart pub repo, but it still works.

Closes #570

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -602,7 +602,7 @@ endfunction
 function! s:CheckFTCtags(bin, ftype) abort
     " dart_ctags isn't installed as a bin
     " must use pub command to run
-    if a:bin ==? 'dart_ctags'
+    if a:bin ==# 'dart_ctags'
         if has('win32')
             let fn = '!pub global list | findstr dart_ctags'
         else
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -151,11 +151,11 @@ function! s:InitTypes() abort
         let type_dart.scope2kind = {
             \ 'class' : 'c'
         \ }
-        let type_dart.ctagsbin   = 'pub'
-        let type_dart.ctagsargs  = 'global run dart_ctags:tags -l'
-        let type_dart.ftype = 'dart'
+        let type_dart.ctagsbin   = dart_ctags
+        let type_dart.ctagsargs  = '-l'
+        let type_dart.ftype      = 'dart'
         call type_dart.createKinddict()
-        let s:known_types.dart = type_dart
+        let s:known_types.dart   = type_dart
     endif
 
     " Use jsctags/doctorjs if available
@@ -603,22 +603,6 @@ endfunction
 
 " s:CheckFTCtags() {{{2
 function! s:CheckFTCtags(bin, ftype) abort
-    " dart_ctags isn't installed as a bin
-    " must use pub command to run
-    if a:bin ==# 'dart_ctags'
-        if has('win32')
-            let fn = '!pub global list | findstr dart_ctags'
-        else
-            let fn = '!pub global list | grep dart_ctags'
-        endif
-        execute(fn)
-        if v:shell_error == 0
-            return a:bin
-        else 
-            return ''
-        endif
-    endif
-
     if executable(a:bin)
         return a:bin
     endif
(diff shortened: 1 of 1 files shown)


jrc2139:
i don't really understand what the `stl` option does.  maybe we change them back to 0?

autoload/tagbar.vim
```diff
@@ -534,6 +567,20 @@ endfunction
 
 " s:CheckFTCtags() {{{2
 function! s:CheckFTCtags(bin, ftype) abort
+    " dart_ctags isn't installed as a bin
+    " must use pub command to run
```
alerque: I don't understand this comment. On which platform would it not be installed as a binary? And what exactly _is_ it? Excuse my ignorance of whatever tech is being employed here

autoload/tagbar.vim
```diff
@@ -534,6 +567,20 @@ endfunction
...
+    if a:bin ==? 'dart_ctags'
```
alerque: Will this ever conceivably be lower case? If its not likely lets just assume that:

```suggestion
    if a:bin ==# 'dart_ctags'
```

alerque:
This PR looks pretty good, thanks for the contribution. If you could clarify for me what's going on with the command execution for my piece of mind it's probably about ready to go. If we have to call some weird command to find out if its there how are we even executing it later? Is everything getting escaped properly?

autoload/tagbar.vim
jrc2139: `pub` is dart's package manager, but it only installs the cli programs to the `pub` path.  so, you have to invoke `pub global run dart_ctags:tags` to run it.  it's possible to clone the repo and compile it into a binary and include that in your `PATH`, but i was considering maximum compatibility/ease of use.

alerque: > so, you have to invoke pub global run dart_ctags:tags to run it

Okay, in that case this PR doesn't seem like it will quite cover things as it just passes the unmodified `a:bin` back as the thing to be executed. You might be using `pub` to _find_ if the specialty provider is installed but I don't see how would actually get executed as that is not part of the path & command that actually get run.

jrc2139: oops I forgot to add the return if it fails.  I figured we don't want to run to generate tags on a random directory, so that's why I did the pub global list command to check if it was installed.

jrc2139:
> This PR looks pretty good, thanks for the contribution. If you could clarify for me what's going on with the command execution for my piece of mind it's probably about ready to go. If we have to call some weird command to find out if its there how are we even executing it later? Is everything getting escaped properly?

ok I've put another [PR](https://github.com/nerdrew/dart-ctags/pull/6) up that will install `dart_ctags` as an executable.  the readme also includes recommended native compilation post installation.  I reverted the CheckFTC function as it was previously.

alerque:
Thanks for the contribution, this looks good now. Let me know if for some reason the upstream change doesn't go through and we need to adjust.

PR #584, opened 2020-01-14
possible bug in AutoUpdate function

### LINKED ISSUE #558: no tags found...but they were previously there
Hi,
this is a description of the odd behavior which I experienced quite frequently
using tagbar:

1) I enable the tagbar for a c++ file (mcsim_boxes.hpp in the logfile attached)
2) all tags of such file are correctly shown in the tagbar
3) I switch to another file in the same buffer (called partcile_lj.hpp in the logfile)
4) I switch back to the file mcsim_boxes.hpp 
5) in the tagbar there are not tags but it says: "no tags found"

this behavior seems to appear after a certain amount
of switches among all files of my project.
A log file, obtained by TabgarDebug command, is attached.

[tagbardebug.log.gz](https://github.com/majutsushi/tagbar/files/3767012/tagbardebug.log.gz)

cridemichel:
this patch should possibly fix #558

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2489,7 +2489,7 @@ function! s:AutoUpdate(fname, force, ...) abort
 
     " Whether we want to skip actually displaying the tags in Tagbar and only
     " update the fileinfo
-    let no_display = a:0 > 0 ? a:1 : 0
+    let no_display = a:0 > 2 ? a:3 : 0
 
     " This file is being loaded due to a quickfix command like vimgrep, so
     " don't process it


alerque:
Thanks for the contribution @cridemichel! I poked around and it seems pretty clear this is indeed a bug. Whether it fixes the other problem or not remains to be seen.

cridemichel:
good to know, I hope it will fix my issue too :-)

cridemichel:
bug #558 unfortunately is still there...

alerque:
Too bad. I've reopened the issue so we can keep track of it. This was still a good catch...

PR #585, opened 2020-01-14
possible bug adding a new filename to _files dictionary

cridemichel:
If I understood correctly the logic, this bug is rather critical since fileinfo is used as dictionary index intead of the filename. Anyway I can not prove it is related to #558.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -663,8 +663,8 @@ endfunction
 " s:known_files.put() {{{2
 " Optional second argument is the filename
 function! s:known_files.put(fileinfo, ...) abort dict
-    if a:0 == 1
-        let self._files[a:1] = a:fileinfo
+    if a:0 == 2
+        let self._files[a:2] = a:fileinfo
     else
         let fname = a:fileinfo.fpath
         let self._files[fname] = a:fileinfo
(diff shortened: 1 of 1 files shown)


alerque:
Again, thanks for the contribution. I squashed this because you submitted it with a previous commit already merged, if you work off an updated master there will be a cleaner history for reviewing PRs ;-)

cridemichel:
> Again, thanks for the contribution. I squashed this because you submitted it with a previous commit already merged, if you work off an updated master there will be a cleaner history for reviewing PRs ;-)

ok I apologize for the inconvenient, next time I will work of an updated master...

cridemichel:
I think we have to revert both commits since a:0 counts the number of variable arguments and *not* the total number of arguments hence previous code both in AutoUpdate and put functions was ok....

alerque:
Oops. Are you sure? I actually looked up the docs on that myself before merging the first PR. I could have read it wrong as well...

cridemichel:
I did some experiments writing some log messages,
but double check before reverting commits…

Cristiano De Michele
cristiano.demichele@gmail.com

alerque:
@cridemichel Reverted both.

cridemichel:
ok

Cristiano De Michele
cristiano.demichele@gmail.com

PR #586, opened 2020-01-15
Only set local settings, not global ones

alerque:
(Found this old branch on my local repo, not sure if it is still relevant.)


zhmars:
According to the document, `ballooneval` is a global option, not global-local. But it can be changed in the future.

alerque:
@zhmars in correct, in VIM this is a global only setting. In NeoVIM it has been removed entirely. This PR seems moot.

PR #589, opened 2020-01-23
ProcessFile should not process an unloaded nor an unexisting buffer

cridemichel:
If buffer is not loaded or does not exist, exit from ProcessFile function, since
in this case getbufline function would return an empty list, thus creating a void file
to be processed for tags extraction. This patch fixes issue #558.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1148,13 +1148,13 @@ function! s:ProcessFile(fname, ftype) abort
         return
     endif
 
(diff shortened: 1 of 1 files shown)


alerque:
Thank you for tracking this down @cridemichel, I know that took a lot of time!

cridemichel:
yes it did but I enjoy so much using it that it was worth the time spent!


Cristiano De Michele
cristiano.demichele@gmail.com

PR #603, opened 2020-05-05
tagbar did not play nicely with popup windows

cridemichel:
open tagbar and after that open the fzf  popup window (I am assuming fzf.vim plugin
is installed), e.g. with

```vim
:History
```
after you have closed the popup window you get the vim error

```text
not allowed to enter a popup window
```

since tagbar attempts to switch to the popup window through the function
goto_win() but this is forbidden. This patch solves this issue by
checking that the winid of the window tagbar is switching to is not a popup window

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3371,6 +3371,18 @@ endfunction
 
 " s:goto_win() {{{2
 function! s:goto_win(winnr, ...) abort
+    "Do not go to a popup window to avoid errors.
+    "Hence, check first if a:winnr is an integer,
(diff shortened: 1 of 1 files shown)


alerque:
FYI in the future you can rebase / squash / amend or whatever to your local branch and force-push to Github to update PRs in place, you don't have to close/reopen them. Also you can use Draft mode to post stuff for CI checking without the risk of somebody coming along and merging it prematurely.

cridemichel:
ok, thank you very much for the information!

Cristiano De Michele
cristiano.demichele@gmail.com

PR #604, opened 2020-05-05
tagbar did not play nicely with popup windows

cridemichel:
steps to reproduce the issue which this PR intends to fix:
 - open tagbar
-  open an fzf popup window (I assume fzf.vim plugin in installed), e.g. use the command:
```vim
:History
```
- close the popup window
- you get the following error:

```text
E366 error not allowed to enter a popup window
``` 
since tagbar attempts to switch to the popup through the function goto_win().
This PR prevents tagbar from switching to popup windows.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3371,6 +3371,18 @@ endfunction
 
 " s:goto_win() {{{2
 function! s:goto_win(winnr, ...) abort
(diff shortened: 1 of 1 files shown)


alerque:
Thanks again for taking the time to track this down and contribute!

cridemichel:
thank you for your prompt reply!

Cristiano De Michele
cristiano.demichele@gmail.com

PR #607, opened 2020-05-13
window local tagbar

isharaka:
adds a global option to allow keeping tagbar local to the window containing the buffer

This is draft pull request.

The intention is to keep the tagbar local to the current window. This change does do that.
However if we move to another window while tagbar is active the tagbar will be updated to reflect the buffer in the active window.

I would like to do the following when switching windows with tagbar active (if the opted in)

prevent tagbar being updated when moving around windows
make TagbarToggle close the active tagbar and reopen it in the newly active window
I am very new to vimscript. I would appreciate some hints to be able to do that.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -888,9 +888,18 @@ function! s:OpenWindow(flags) abort
         let openpos = g:tagbar_left ? 'leftabove ' : 'rightbelow '
         let width = g:tagbar_vertical
     endif
+
+    let l:splitright = &splitright
+    if g:tagbar_window_local == 1
+        let openpos = ''
+        let &splitright = g:tagbar_left ? 0 : 1
+    endif
+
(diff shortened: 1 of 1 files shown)


alerque:
Some reason this was closed? It it isn't ready but you still plan to work on it it would be nice to track it as an open PR in Draft mode (you can convert it back using the link in the reviewer section after you open it up again).

isharaka:
Hi @alerque, This one is still open. I did open one before and unintentionally closed it.

raven42:
@isharaka A little while ago there was a change made to the `g:tagbar_left` to use `g:tagbar_position` instead (#630 / #652). This allows more granular control of the window positioning so it should now be possible to make the tagbar window local to the active window by doing something like this:
```
let g:tagbar_position = leftabove
```
This should open the tagbar to the left or above of the current active window (see `:h split` for more details on exactly how this works). Note that this wouldn't cover the case of only updating the tagbar window when switching to other windows. It will still do that, but this should at least give you a few more options for window positioning.

Issue #477, opened 2018-02-03
Avoid tagbar runs all the time in new opened buffer but on demand


niva-xx:
Hi,

From usage section in help, tagbar is running all the time and each time I open a file in buffer.
It pass by vimrun out of job_start( to run ctag executable and update tags I imagine.

How ca I set option in order to disable this 'all the time' mode and switch to 'on demand' mode ?

Thank you

skywind3000:
background ctags calling costs me 5-6s each time on windows.

https://github.com/majutsushi/tagbar/issues/475

niva-xx:
Hum OK i understand but is there an option to disable 'always' ctags launch
on each opened buffer and launch it on demand.

skywind3000:
There is an alternative for tagbar, currently I have switch to LeaderF/CtrlP from tagbar. LeaderF/CtrlP can display function list of the current buffer and apply fuzzy match on the function list:

In LeaderF you can use command `:LeaderfFunction` (bound to my F9) to display a function list like tagbar:

![leaderf2](https://user-images.githubusercontent.com/3035071/35934809-86eb76c0-0c79-11e8-896b-8e519af092da.png)

ctags are invoked on demand and executed in the background. You can use c-j, c-k to navigate between function names or use leaderf's fuzzy matcher to find one.

The only disadvantage of leaderf is python requirement. If you don't get the python support in your vim, you can also use [CtrlP](https://github.com/ctrlpvim/ctrlp.vim) which is definitely 100% vimscript.

CtrlP has a plugin [ctrlp-funky](https://github.com/tacahiroy/ctrlp-funky) which provide a `:CtrlPFunky` command for you to display the function list like LeaderF and enable you using arrow keys or fuzzy match to navigate function names.

Now I am very happy with them.

niva-xx:
Hi and thank you a lot for this alternative.

CtrpFunky seems fast. But It does not display class with public/private
methods like tagbar isn't it ?

Best,
Nicholas

2018-02-07 19:49 GMT+01:00 Linwei <notifications@github.com>:

skywind3000:
Another promising competitor you may have a look:  [vista](https://github.com/liuchengxu/vista.vim)

keywords: universal-ctags, lsp, background task, popup-window

alerque:
I think fixing this is urgent. Most of the time it doesn't bother me, but opening a big file somewhere and not saving it shouldn't result in a couple seconds of hand and a tags file I never used!

[...]

PR #611, opened 2020-05-29
windows: prevent temporary window creation

skywind3000:
New version of: https://github.com/majutsushi/tagbar/pull/348

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2868,7 +2868,7 @@ endfunction
 
 " run shell command in a proper way: prevent temporary window creation
 function! s:run_system(cmd, version) abort
-    if has('win32') && (!has('nvim')) && (has('python3') || has('python2')) && a:version > 0
+    if has('win32') && (!has('nvim')) && a:version > 0 && (has('python3') || has('python2'))
         if a:version == 3 && has('python3')
             let pyx = 'py3 '
             let python_eval = 'py3eval'
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2866,6 +2866,45 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" python simulate system() on window to prevent temporary window creation
+function! s:system_python(cmd, version)
+    if has('nvim')
+        let hr = system(a:cmd)
+    elseif has('win32') || has('win64') || has('win95') || has('win16')
+        if a:version < 0 || (has('python3') == 0 && has('python2') == 0)
+            let hr = system(a:cmd)
+            let s:shell_error = v:shell_error
+            return hr
+        elseif a:version == 3
+            let pyx = 'py3 '
+            let python_eval = 'py3eval'
+        elseif a:version == 2
+            let pyx = 'py2 '
+            let python_eval = 'pyeval'
+        else
+            let pyx = 'pyx '
+            let python_eval = 'pyxeval'
+        endif
+        exec pyx . 'import subprocess, vim'
+        exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
+        exec pyx . '__argv["stdout"] = subprocess.PIPE'
+        exec pyx . '__argv["stderr"] = subprocess.STDOUT'
+        exec pyx . '__pp = subprocess.Popen(**__argv)'
+        exec pyx . '__return_text = __pp.stdout.read()'
+        exec pyx . '__pp.stdout.close()'
+        exec pyx . '__return_code = __pp.wait()'
+        exec 'let l:hr = '. python_eval .'("__return_text")'
+        exec 'let l:pc = '. python_eval .'("__return_code")'
+        let s:shell_error = l:pc
+        return l:hr
+    else
+        let hr = system(a:cmd)
+    endif
+    let s:shell_error = v:shell_error
+    return hr
+endfunction
+
+
 " s:ExecuteCtags() {{{2
 " Execute ctags with necessary shell settings
 " Partially based on the discussion at
@@ -2903,7 +2942,8 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        silent let ctags_output = system(a:ctags_cmd)
+        let py_version = get(g:, 'tagbar_python', 0)
+        silent let ctags_output = s:system_python(a:ctags_cmd, py_version)
     endif
 
     if &shell =~? 'cmd\.exe'
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2868,8 +2868,7 @@ endfunction
 
 " run shell command in a proper way: prevent temporary window creation
 function! s:run_system(cmd, version) abort
-    if (has('win32') || has('win64')) && has('nvim') == 0
-        if a:version <= 0 || (has('python3') == 0 && has('python2') == 0)
+    if !has('nvim') && has('win32') && a:version >= 0 && (has('python3') || has('python2'))
             let hr = system(a:cmd)
             let s:shell_error = v:shell_error
             return hr
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2868,7 +2868,7 @@ endfunction
 
 " run shell command in a proper way: prevent temporary window creation
 function! s:run_system(cmd, version) abort
-    if has('win32') && (!has('nvim')) && a:version > 0 && (has('python3') || has('python2'))
+    if has('win32') && !has('nvim') && a:version > 0 && (has('python3') || has('python2'))
         if a:version == 3 && has('python3')
             let pyx = 'py3 '
             let python_eval = 'py3eval'
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2892,9 +2892,8 @@ function! s:run_system(cmd, version) abort
         exec 'let l:pc = '. python_eval .'("__return_code")'
         let s:shell_error = l:pc
         return l:hr
-    else
-        let hr = system(a:cmd)
     endif
+    let hr = system(a:cmd)
     let s:shell_error = v:shell_error
     return hr
 endfunction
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for fixing this up and getting it ready to include!

@wsdjeg Would you happen to be available to give this a review? I'll test it to make sure the Linux side breaks but it would be nice to have 3rd party confirmation that Windows works as expected too before I merge.

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" python simulate system() on window to prevent temporary window creation
+function! s:system_python(cmd, version) abort
+    if has('nvim')
```
alerque, line 2871: Does Neovim not have this problem on Windows?

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
...
+        let hr = system(a:cmd)
+    elseif has('win32') || has('win64') || has('win95') || has('win16')
```
alerque, line 2873: Please re-arrange this if/elseif/else structure to just be if/else. There are two separate cases (the current *if* and the *else* that both have the same outcome, only the *elseif* is different. Just make that the primary test and *else* the other option.

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" python simulate system() on window to prevent temporary window creation
+function! s:system_python(cmd, version) abort
```
alerque, line 2870: Lets rename this function to `s:run_system()` or something like that since it may or may not actually execute Python at all.

autoload/tagbar.vim
skywind3000: yes, neovim doesn't need this.

autoload/tagbar.vim
skywind3000: This may look like:

```VimL
    if (has('win32') || has('win64') || has('win95') || has('win16')) && has('nvim') == 0
```

, which makes it a little ugly and exceeding 78 characters in one line.

autoload/tagbar.vim
skywind3000: done

autoload/tagbar.vim
alerque: Most of these are deprecated flags anyway. Just *win32* should be fine for this purpose. Also there is duplicated code in the version==0 case here, and the has() flags for python versions are never actually _used_.

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
...
+        let hr = system(a:cmd)
+    if has('win32') || has('win64') || has('win95') || has('win16')
+        if a:version <= 0 || (has('python3') == 0 && has('python2') == 0)
+            let hr = system(a:cmd)
+            let s:shell_error = v:shell_error
+            return hr
+        elseif a:version == 3
+            let pyx = 'py3 '
+            let python_eval = 'py3eval'
+        elseif a:version == 2
```
alerque, line 2879:
```suggestion
        elseif a:version == 2 || has('python2')
```

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
...
+            let pyx = 'py2 '
+            let python_eval = 'pyeval'
+        else
```
alerque, line 2882: What is this `pyxeval` thing? What are we even targeting here?

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,46 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" run shell command in a proper way: prevent temporary window creation
+function! s:run_system(cmd, version) abort
+    if has('nvim')
+        let hr = system(a:cmd)
+    if has('win32') || has('win64') || has('win95') || has('win16')
+        if a:version <= 0 || (has('python3') == 0 && has('python2') == 0)
```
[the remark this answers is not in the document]
alerque: I believe *win64* is a subset of *win32*, which actually covers both 32 and 64 bit systems. We don't need to test for it separately.

autoload/tagbar.vim
skywind3000: the has('python') is necessary if `a:version` is above zero but user's vim is compiled without python.

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,44 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" run shell command in a proper way: prevent temporary window creation
+function! s:run_system(cmd, version) abort
+    if (has('win32') || has('win64')) && has('nvim') == 0
+        if a:version <= 0 || (has('python3') == 0 && has('python2') == 0)
```
alerque, line 2872:
```suggestion
    if !has('nvim') && has('win32') && a:version >=0 && (has('python3') || has('python2'))
```

autoload/tagbar.vim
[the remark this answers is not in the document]
skywind3000: see `:h feature-list`

```
win16			old version for MS-Windows 3.1 (always false)
win32			Win32 version of Vim (MS-Windows 95 and later, 32 or
			64 bits)
win32unix		Win32 version of Vim, using Unix files (Cygwin)
win64			Win64 version of Vim (MS-Windows 64 bit).
win95			Win32 version for MS-Windows 95/98/ME (always false)
```

To be safe, I have changed it to:

```VimL
    if (has('win32') || has('win64')) && has('nvim') == 0
        ...
    else
        ...
    endif
```

autoload/tagbar.vim
skywind3000: the purpose of `a:version` is for user to indicate python version:

- 0: disable this feature
- 1: automatically choose python version, by ultilizing `pyx` and `pyxeval` commands.
- 2: force use python 2 
- 3: force use python 3

Both `pyx` and `pyxeval` are internal vim commands to perform `py`/`pyeval` or `py3`/`py3eval` automatically.

autoload/tagbar.vim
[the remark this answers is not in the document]
alerque: Please don't "be safe" by including code that is documented to be un-necessary. Modern 64 bit VIMs have always supported `has('win32')` and at this point always will. Everywhere else in the plugin we are only checking that flag, I'd rather not have inconsintent code or platform detection.

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,43 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
...
+        exec pyx . '__return_text = __pp.stdout.read()'
+        exec pyx . '__pp.stdout.close()'
+        exec pyx . '__return_code = __pp.wait()'
+        exec 'let l:hr = '. python_eval .'("__return_text")'
+        exec 'let l:pc = '. python_eval .'("__return_code")'
+        let s:shell_error = l:pc
+        return l:hr
+    else
+        let hr = system(a:cmd)
+    endif
```
alerque, line 2900:
```suggestion
    endif
    let hr = system(a:cmd)
```

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,43 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" run shell command in a proper way: prevent temporary window creation
+function! s:run_system(cmd, version) abort
+    if !has('nvim') && has('win32') && a:version >= 0 && (has('python3') || has('python2'))
+            let hr = system(a:cmd)
+            let s:shell_error = v:shell_error
+            return hr
+        elseif a:version == 3
```
alerque, line 2875:
```suggestion
        if a:version == 3 && has('python3')
```

autoload/tagbar.vim
alerque: Nevermind I didn't realize that cal a call() thing, carry on...

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,44 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
...
 endfunction
 
+" run shell command in a proper way: prevent temporary window creation
+function! s:run_system(cmd, version) abort
+    if (has('win32') || has('win64')) && has('nvim') == 0
+        if a:version <= 0 || (has('python3') == 0 && has('python2') == 0)
+            let hr = system(a:cmd)
+            let s:shell_error = v:shell_error
+            return hr
+        elseif a:version == 3 && has('python3')
```
alerque:
```suggestion
    if !has('nvim') && a:version >= 0 && has('win32') && (has('python3') || has('python2') == 0))
        if a:version == 3 && has('python3')
```

autoload/tagbar.vim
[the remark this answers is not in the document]
skywind3000: updated

autoload/tagbar.vim
```diff
@@ -2866,6 +2866,39 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
     return ctags_cmd
 endfunction
 
+" run shell command in a proper way: prevent temporary window creation
+function! s:run_system(cmd, version) abort
+    if has('win32') && (!has('nvim')) && a:version > 0 && (has('python3') || has('python2'))
```
alerque, line 2871:
```suggestion
    if has('win32') && !has('nvim') && a:version > 0 && (has('python3') || has('python2'))
```

Almost there!

skywind3000:
@wsdjeg may help.

zhmars:
Not sure which is better, Job or the Python method. 

How about using `job_start` when it is available?

alerque:
Thanks @zhmars. I think that using job control is the long term best way to address #477, but that's a slightly different issue. This PR is more of a stop gap to avoid a bad behavior on Windows, proper job control and async background processing can probably be left to another PR.

Were you able to confirm that this actually works and Windows and doesn't break anything serious?

PR #616, opened 2020-06-01
Fix CI runs to actually run nvim instead of just claiming to

alerque:


alerque:
Besides the egregious error in my testing setup that was getting all set to run nvim then executing vim instead, the Actions environment seems to have changed because ctags doesn't run at all now. Worse, all the tests are passing even though none of them run. :frowning:

PR #625, opened 2020-06-21
Reset window after jump to tag if zoomed

tanhuacheng:


alerque:
Thanks for taking the time to contribute.

Can you walk me through what scenario this is useful in? I don't even know how to test whether it works properly. How do I get into the situation where this would change the behavior?

tanhuacheng:
> Thanks for taking the time to contribute.
> 
> Can you walk me through what scenario this is useful in? I don't even know how to test whether it works properly. How do I get into the situation where this would change the behavior?

When the tagbar window is zoomed, the tagbar window is still zoomed after jump to  tag, we still need to  switching to the tagbar window and then restor it's normal size everytime.

## befor jump to tag
image
## befor this patch & after jump to tag
image
## after this path & after jump to tag
image

PR #630, opened 2020-07-27
Add more flexible options for window positioning.

raven42:
New configuration option for g:tagbar_position that will supercede g:tagbar_left. This option allows for new values for 'top', 'bottom', 'left' and 'right' (default:'right') to position the Tagbar window.

New configuration option for g:tagbar_height that will supercede g:tagbar_vertical. This option is used to set the Tagbar window height when the window is split using the 'top' or 'bottom' position.

New configuration option for g:tagbar_no_status_line. This option will disable any Tagbar status line updates in the event another plugin is controlling the status line. This prevents Tagbar from changing the status line when Tagbar is not the active window.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -879,14 +879,21 @@ function! s:OpenWindow(flags) abort
     endif
 
     let s:window_opening = 1
-    if g:tagbar_vertical == 0
+    if g:tagbar_position =~# '\v(bottom|right)'
+        let openpos = 'rightbelow '
+    else
+        let openpos = 'leftabove '
+    endif
+    if g:tagbar_position =~# '\v(left|right)'
         let mode = 'vertical '
-        let openpos = g:tagbar_left ? 'topleft ' : 'botright '
         let width = g:tagbar_width
     else
         let mode = ''
-        let openpos = g:tagbar_left ? 'leftabove ' : 'rightbelow '
-        let width = g:tagbar_vertical
+        if g:tagbar_height > 0
+            let width = g:tagbar_height
+        else
+            let width = g:tagbar_vertical
+        endif
     endif
     exe 'silent keepalt ' . openpos . mode . width . 'split ' . s:TagbarBufName()
     exe 'silent ' . mode . 'resize ' . width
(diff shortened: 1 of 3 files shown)


autoload/tagbar.vim
```diff
@@ -879,14 +879,21 @@ function! s:OpenWindow(flags) abort
...
 
     let s:window_opening = 1
-    if g:tagbar_vertical == 0
+    if g:tagbar_position =~# '\v(bottom|right)'
+        let openpos = 'rightbelow '
+    else
+        let openpos = 'leftabove '
+    endif
+    if g:tagbar_position =~# '\v(left|right)'
         let mode = 'vertical '
```
alerque, line 888: I understand this isn't part of the code you added, but something doesn't sit right with me here. Why is this defined with a space at the end of the variable's contents? Why is the other option '' (instoad of 'horizontal')?

This PR looks great over-all and is an improvement, but I'm wondering if perhaps there is a little bit more that could be cleaned up at the same time (separate commit maybe, but same PR). Cat we make even these local variables sane while we're mucking with this?

raven42: My intent was to leave the commands being executed the same as much as possible. The extra spaces at the end are because they are all being concatenated together below at the `exe 'silent keepalt ' . openpos . mode . width . 'split ' . sTagbarBufName()` line. If you have any suggestions for how you'd like this cleaned up, I can certainly take a shot at it.

raven42: Also in reference to other option being '' instead of 'horizontal', that is because of the `split` command being executed. In vim, there is no `split horizontal` command. It is either `split` or `split vertical`. This code is constructing a command to be executed in vim to split the window according to the requested parameters.

alerque: Thanks for the feedback on this. I see that's kind of a hacky way to make this happen but understand why it is the way it is. I think for now you might be right, the best thing is to just leave it alone.

alerque:
Thanks for the contribution @raven42!

marceldev89:
This PR introduces screen flashing/redrawing when using the `tagbar#currenttag()` function in the status line. Going back to the commit before the merge it's fine again.

raven42:
Have a pull-request up to fix this. There was a redraw incorrectly left in during this pull request when I was looking at a character encoding issue.
https://github.com/majutsushi/tagbar/pull/634

PR #633, opened 2020-07-31
Add Update() function to trigger an AutoUpdate() of the current file

raven42:


alerque:
Thanks for taking the time to contribute!

Can I ask what the use case for this is? I've never needed to manually update, and if `AutoUpdate` was really 'automatic' it seems like this shouldn't be necessary. I'm not opposing this, I'd just like to understand under what circumstances it's needed.

raven42:
This specific issue I'm seeing is integrating multiple plugins together into a common windowed environment. I am using NERDTree and Tagbar at the same time, and there is an issue I'm seeing when opening these windows automatically where I can get the placement right, and I've added some updates to my .vimrc to activate the correct window after loading, however when I am activating that window, the tagbar window is not refreshing itself. It only refreshed upon the cursor movement, and not activation of the window.

The driving force behind this is likely an issue I'm facing with my work dev environment. The software packages we are using are extremely old and I'm seeing some odd issues.

The basic window setup looks like this:
```
+----------+---------------+
| NERDTree | Edit Window   |
|          |               |
|          |               |
+----------+               |
| Tagbar   |               |
|          |               |
|          |               |
+----------+---------------+
```
My .vimrc for the relevant portion looks like this with the new function being called in the ActivateWindow function:
```
" --- https://github.com/majutsushi/tagbar
let g:have_tagbar = &runtimepath =~ 'tagbar' ? 1 : 0
" --- https://github.com/preservim/nerdtree
let g:have_nerdtree = &runtimepath =~ 'nerdtree' ? 1 : 0

let g:tagbar_position = g:have_nerdtree ? 'bottom' : 'left'
let g:tagbar_height = winheight(0) / 2
let g:tagbar_width = winwidth(0) > 150 ? 50 : winwidth(0) / 3
let g:tagbar_previewwin_pos = 'botright'

autocmd VimEnter *.c,*.cpp,*.h,*.py call devpanel#DevPanelOpen()

function! devpanel#DevPanelOpen()
	if winwidth(0) < 100
		return
	endif
	if (!exists('t:dev_panel_open'))
		let t:dev_panel_open = 0
	endif
	if (t:dev_panel_open == 0)
		set number
		let t:dev_panel_open = 1
		call devpanel#DevPanelMarkWindow()
		call devpanel#DevPanelSizeUpdate()
		if g:have_nerdtree
			silent NERDTree %
		endif
		if g:have_tagbar
			silent TagbarOpen
		endif
		call devpanel#DevPanelActivateMarkedWindow()
	endif
endfunction

function! devpanel#DevPanelMarkWindow()
	let w:devpanel_marked_window = 1
endfunction

function! devpanel#DevPanelActivateMarkedWindow()
	for window in range(1, winnr('#'))
		execute window . 'wincmd w'
		if exists('w:devpanel_marked_window')
			unlet w:devpanel_marked_window
			if g:have_tagbar
				call tagbar#Update()                           " <- Update Tagbar here when activating edit window
			endif
			return
		endif
	endfor
endfunction

function! devpanel#DevPanelSizeUpdate()
	if g:have_nerdtree
		if g:have_tagbar
			"let g:tagbar_vertical = winheight(0) / 2
			let g:tagbar_height = winheight(0) / 2
		endif
		let g:NERDTreeWinSize = winwidth(0) > 150 ? 50 : winwidth(0) / 3
	endif
	redraw!
endfunction

```

alerque:
Thanks for the detailed explanation. I'm pretty sure this isn't even a bug in Tagbar and what you are working around is *other* issues, but I also don't see any harm in making this function available for anybody else in a similar position.

raven42:
Thanks. Ya I also didn't think it was anything to do with Tagbar itself. Likely the older tools I'm working with, so rather than changing any autocmds with Tagbar itself, I thought an Update() function would help me work around my issue without impacting any other functionality.

PR #634, opened 2020-08-03
Fix the screen flicker by removing the redraw without debug enabled. …

raven42:
…Mistakenly left in during commit 4e7e199


alerque:
Thanks for fixing this.

PR #639, opened 2020-08-08
Correct last contribution that contained an syntax error

embear:
Sadly my last contribution contained an syntax error. I don't understand how this happened. Sorry!


alerque:
Gosh I thought I merged this already. Apparently I only pulled it locally. Sorry about that, and thanks for the fix!

PR #642, opened 2020-08-15
Fix maximized state track on close

traysh:
Fix need of pressing 'x' twice to maximize the Tagbar when it was closed maximized.


alerque:
Thanks for taking the time to contribute this fix @traysh.

PR #643, opened 2020-08-18
Add user facing function to get tag near line number

raven42:
Add a user facing function to get the tag near a specified line number. Use case would be to use this in a custom foldtext function so the tag can be displayed as part of the fold text.


alerque:
Thanks for the contribution @raven42!

PR #631, opened 2020-07-30
Error when using tagbar#currenttagtype() with Rust

embear:
I'm using `tagbar#currenttagtype()` to display the tag type in my status line. This works fine for C and C++ but fails with Rust. It seems that all `typeinfo.kinds` of universal ctags for Rust are already singular so getting the singular expression from the `s:singular_types` dictionary fails. As a fix I'm checking if the key is available in the dictionary before accessing it.

System information:

  * VIM - Vi IMproved 8.2 (2019 Dec 12, compiled May 18 2020 10:06:50) Included patches: 1-788
  * Universal Ctags 0.0.0(3f0ea94c)

For reference I printed the content of `typeinfo.kinds`:

```
[{'short': 'n', 'long': 'module', 'fold': 1, 'stl': 0}, {'short': 's', 'long': 'struct', 'fold': 0, 'stl': 1}, {'short': 'i', 'long': 'trait', 'fold': 0, 'stl': 1}, {'short': 'c', 'long': 'implementation', 'fold': 0, 'stl': 0}, {'short': 'f', 'long': 'function', 'fold': 0, 'stl': 1}, {'short': 'g', 'long': 'enum', 'fold': 0, 'stl': 1}, {'short': 't', 'long': 'type alias', 'fold': 0, 'stl': 1}, {'short': 'v', 'long': 'global variable', 'fold': 0, 'stl': 1}, {'short': 'M', 'long': 'macro', 'fold': 0, 'stl': 1}, {'short': 'm', 'long': 'struct field', 'fold': 0, 'stl': 1}, {'short': 'e', 'long': 'enum variant', 'fold': 0, 'stl': 1}, {'short': 'P', 'long': 'method', 'fold': 0, 'stl': 1}, {'short': '?', 'long': 'unknown', 'fold': 0, 'stl': 1}]
```

PR #645, opened 2020-08-23
Fix has_key() parameters in  currenttagtype function.

### LINKED ISSUE #644: currenttagtype complains about usage of has_keys.
I'm experiencing issues with latest tagbar. I have a `currenttag` and `currenttagtype` on my status bar, after opening vim with a file (or opening it using `:e`), I have immediately this error:

```
Error detected while processing function tagbar#currenttagtype:
line   18:
E119: Not enough arguments for function: has_key
```

gryf:
Closes #644.

The idea of unconditionally reaching for particular word in dictionary using plural key was, that in case of missing singular, it vim will complain about it, so that it might be added to the dictionary in next iteration. Currently, it will offer plural as singular which could be easily missed.

Maybe behaviour for particular ctags implementations should be separated, as apparently that was the case in PR #631?

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3720,7 +3720,7 @@ function! tagbar#currenttagtype(fmt, default) abort
 
     let typeinfo = tag.fileinfo.typeinfo
     let plural = typeinfo.kinds[typeinfo.kinddict[kind]].long
-    if has_key(plural)
+    if has_key(s:singular_types, plural)
         let singular = s:singular_types[plural]
     else
         let singular = plural


alerque (approved):

alerque:
Thanks for taking the time to track this down and fix it. In this case I'm not sure how separating the code paths per ctags version would help, this seems to be a straight up coding bug. But in general yes I agree we should probably work on a cleaner separation of what functions we call in our workflow vs. how those calls are actually make on the system and how the data is normalized coming back. It's quite a bit of a hodge-podge right now.

PR #649, opened 2020-08-29
Fix issue with jumping to a tag with special characters

bravestarr:
In C++, the name of the destructor of a class starts with a tidle(~) which is special when the option 'magic' is set. This will cause an error when jumping to a destructor. Calling search() in "very nomagic" environment can solve this problem no matter the option 'magic' is set or not.


alerque:
Thanks for tracking this down and cotributing a fix @bravestarr, LGTM.

alerque (approved):

PR #652, opened 2020-08-31
Fixes #650 Correct window positioning defaults

### LINKED ISSUE #650: How to get back the old default (botright vertical) window position?
My vimscript is a little rusty so I may be missing something obvious here.  Can i get my normal `botright vertical` tagbar window position back somehow?

I was always happy with the default window position for tagbar and never changed it in my .vimrc.   It seems to me that since #630 the default position is now `rightbelow verical` (i.e. right of the current window when opening tagbar) rather than `botright vertical`, as it was always before, (i.e. far right of the vim window).

From a quick glance at the help for `g:tagbar_position` and 4e7e1990429a4b4d7191719c40b23d9a084c237a I can't see a way to get back the original default behavior using settings in my `.vimrc`.  But as I said it's been a while since I looked at any vimscript so forgive me if I've missed something obvious.

(For now I've reverted to before 4e7e1990429a4b4d7191719c40b23d9a084c237a.)

raven42:
Fixes #650 
With default case having multiple vertically split windows, the tagbar
window should open on the far right of the screen. When the feature in
pull-request #630 was added, this behavior changed to open the tagbar
window relative to the current window. This commit will fix that by
allowing direct usage of the vim split options.

Validated the following cases for the old default values. These all
still behave the same way as prior to pull-request #630 when the tagbar
is activated in the `active` window shown in the below examples:
```
	" No options - tagbar on right, preview above
	+-----------------------------------+
	| preview                           |
	+--------+--------+--------+--------+
	| edit   | edit   | edit   | tagbar |
	|        | active |        |        |
	+--------+--------+--------+--------+

	" tagbar on left, preview above
	let g:tagbar_left = 1
	+-----------------------------------+
	| preview                           |
	+--------+--------+--------+--------+
	| tagbar | edit   | edit   | edit   |
	|        |        | active |        |
	+--------+--------+--------+--------+

	" tagbar on bottom, preview to right of tagbar
	let g:tagbar_vertical = 10
	+--------+---------------+--------+
	| edit   | edit          | edit   |
	|        | active        |        |
	|        +-----+---------+        |
	|        | tag | preview |        |
	+--------+-----+---------+--------+

	" tagbar on top, preview to the right of tagbar
	let g:tagbar_left = 1
	let g:tagbar_vertical = 10
	+--------+-----+---------+--------+
	| edit   | tag | preview | edit   |
	|        +-----+---------+        |
	|        | edit          |        |
	|        | active        |        |
	+--------+---------------+--------+

```

New values supported by #630 also all validated using the defined behavior
```
let g:tagbar_position = 'right'   " Behaves the same as no options
let g:tagbar_position = 'left'    " Behaves the same as g:tagbar_left = 1
let g:tagbar_position = 'bottom'  " Behaves the same as g:tagbar_vertical = <#>
let g:tagbar_position = 'top'     " Behaves the same as g:tagbar_left = 1 g:tagbar_vertical = <#>
```

### DIFF
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -446,24 +446,26 @@ configuration files.
 
                                                            *g:tagbar_position*
 g:tagbar_position~
-Default: 'right'
+Default: 'botright vertical'
 
-By default the Tagbar window will be opened on the right-hand side of vim. Set
-this option to one of 'left', 'right', 'bottom', or 'top' to open in the
-corresponding position instead. This can be useful when activating Tagbar at
-the same time as another plugin which creates a new window. It allows for more
-granular control of the Tagbar position in relation to the current active
-window.
+By default the Tagbar window will be opened on the right-hand side of vim in a
+vertical split. Set this option to one of the standart vim split options such
+as 'topleft', 'botright', 'leftabove', or 'rightbelow' to open in the
+corresponding position instead. If desiring a vertically split window, then
+include a ' vertical' in the field value as well. This can be useful when
+activating Tagbar at the same time as another plugin which creates a new
+window. It allows for more granular control of the Tagbar position in
+relation to the current active window.
 
-If set to 'top' of 'bottom', |g:tagbar_height| will be used to determine the
-window height for the tagbar window.
+If using a vertical split, |g:tagbar_width| will be used to determine the
+window width for the tagbar window. Else |g:tagbar_height| will be used to
+determine the window height for the tagbar window.
 
-if set to 'left' or 'right', |g:tagbar_width| will be used to determine the
-window width for the tagbar window.
+See |split| for more details on window positioning.
 
 Example:
 >
-        let g:tagbar_position = 'left'
+        let g:tagbar_position = 'leftabove'
 <
                                                                *g:tagbar_left*
 g:tagbar_left~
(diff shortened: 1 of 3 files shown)


alerque:
Thanks @raven42, this is lookig good. I haven't actually had a chance to test drive it but the code looks sane.

@RoDuth have you tried this out yet in practice?

doc/tagbar.txt
```diff
@@ -505,20 +507,20 @@ Example:
 g:tagbar_height~
 Default: 0
 
-If |g:tagbar_position| is set to 'bottom' or 'top', then this value is used to
-determine the height of the Tagbar window.
-See |g:tagbar_left| for configuring the position of the window.
+If |g:tagbar_position| is does not include a 'vertical' option, then this
```
RoDuth, line 510: very minor typo here...
```
If |g:tagbar_position| does not include a 'vertical' option, then this
```
(remove is)

RoDuth:
Thanks for your effort on this @raven42, works for me.

@alerque  I've given it a quick run around the block (behaves as expected), and agree that the code looks good to me, happy for you to merge it.

doc/tagbar.txt
alerque: Please apply this suggestion and I'll merge --- the rest looks good to me.

alerque (changes requested):

doc/tagbar.txt
raven42: fixed

Issue #636, opened 2020-08-04
Is there a way to disable tagbar for very large files


ericsunplus:
Hi there,

I am using tagbar for a long time and enjoy its' feature. However, recently I am profiling a problem that my vim runs very slow when loading a large header file, it turns out the root cause is tagbar makes a lot of iteration for this header file and spend quite a long time. The following is from my profiling result

```
FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
    1  79.733105   0.000300  dist#ft#FTheader()
    4  79.654429   0.021717  <SNR>131_AutoUpdate()
    1  79.631572   4.059338  <SNR>131_ProcessFile()
69277  52.977399  10.519105  <SNR>131_ParseTagline()
69277  42.458294  10.486403  <SNR>131_ProcessTag()
  323  22.706893  22.026039  tagbar#sorting#sort()
69277  22.674011   2.594087  tagbar#prototypes#normaltag#new()
    1  21.794226   0.000258  <SNR>144_sortTags()
69383  18.492879  10.969571  tagbar#prototypes#basetag#new()
862470  17.951957             <SNR>149_compare_by_kind()
75180   8.094827   6.071874  <SNR>131_add_tag_recursive()
1248894   7.523308             <SNR>146_add_snr()
69383   1.849420   1.628480  <SNR>146_initFoldState()
277108   1.614862             <SNR>145_add_snr()
63871   0.819722             <SNR>144_addTag()
    1   0.798148   0.074979  <SNR>131_ExecuteCtagsOnFile()
    1   0.722627   0.034371  <SNR>131_ExecuteCtags()
69277   0.255400             <SNR>144_getTagsByName()
69388   0.220967             263()
69277   0.090358             <SNR>146_isSplitTag()

FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
  323  22.706893  22.026039  tagbar#sorting#sort()
862470             17.951957  <SNR>149_compare_by_kind()
69383  18.492879  10.969571  tagbar#prototypes#basetag#new()
69277  52.977399  10.519105  <SNR>131_ParseTagline()
69277  42.458294  10.486403  <SNR>131_ProcessTag()
1248894              7.523308  <SNR>146_add_snr()
75180   8.094827   6.071874  <SNR>131_add_tag_recursive()
    1  79.631572   4.059338  <SNR>131_ProcessFile()
69277  22.674011   2.594087  tagbar#prototypes#normaltag#new()
[...]

PR #656, opened 2020-09-14
Add option to not trigger update on large files

raven42:
New feature for #636 

Initial addition of large file support option. Just getting the ball rolling on this, but not fully implemented.

Looking to find a way to add output into tagbar window indicating that the file was ignored. Currently this will be displayed once in the vim messages history, but due to ordering of loading, the message isn't actually seen. I believe this is due to the messages history getting initialized after the plugins are loaded.

This adds an option `g:tagbar_file_size_limit` which can be set by the user to set a line count limit so the auto-update does not process anything on the file. This can be overridden by calling a new `tagbar#ForceUpdate()` routine

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1158,22 +1158,6 @@ function! s:ProcessFile(fname, ftype) abort
         return
     endif
 
-    " If the file size limit it set, then check the file size to see if
-    " this file should be ignored or not.
-    if g:tagbar_file_size_limit > 0
-        let fsize = getfsize(expand('%'))
-        if fsize > g:tagbar_file_size_limit && !exists('b:tagbar_force_update')
-            call tagbar#debug#log('[ProcessFile] File size too large (' . fsize .
-                        \ ' bytes) - limit set to ' . g:tagbar_file_size_limit)
-            if !exists('b:tagbar_file_exceeds_limit')
-                echom 'File size too large (' . fsize .
-                            \ ' bytes) - Not processing file (see help for g:tagbar_file_size_limit).'
-                let b:tagbar_file_exceeds_limit = 1
-            endif
-            return
-        endif
-    endif
-
     let typeinfo = s:known_types[a:ftype]
 
     " If the file has only been updated preserve the fold states, otherwise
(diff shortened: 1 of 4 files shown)


autoload/tagbar.vim
```diff
@@ -1158,6 +1158,20 @@ function! s:ProcessFile(fname, ftype) abort
         return
     endif
 
+    " If the file size limit it set, then check the linecount to see if
```
alerque, line 1161: I like this feature, but in reviewing the code I found it very counter intuitive that the thing being checked is lines, but the setting here is called filesize. File size in my mind is a byte length, not a line count.

One of the most common cases I'd like to exclude is "compiled" source files that have been compressed by removing spaces. These show up a lot in JavaScript work and are usually massive rollups of other code jammed into just a few lines. They would easily be over a reasonable limit for filesize, but not trip a line count limit.

I think either this setting should be renamed to be intuitive or changed so it actually works on filesize. I think I'd prefer the latter. What do you think?

alerque (changes requested):

autoload/tagbar.vim
raven42: Updated to work on file byte size instead.

raven42:
Any idea how to display some info in the tagbar window for this?

alerque (dismissed):

alerque:
Thanks for working on this @raven42.

> Anybody have any idea how to display some info in the tagbar window for this?

Something would definitely make for a better UX than nothing.

This shouldn't be an error or anything else disruptive to a workflow, but something stuffed in the Tagbar buffer indicating that the file was deliberately not processed would both disambiguate intended from unintended function and give people a clue what to change to get a different result.

alerque:
Just FYI @raven42 I didn't merge this yet because I wasn't sure if you were going to continue on poking at a status report of some kind. If you're not going to get around to that soon feel free to go ahead and merge when you think it's ready. As long as it has at least one current review approval from anybody else with write access you should me able to merge it yourself.

raven42:
@alerque: Thanks, ya I'm not quite ready to merge it in. Was going to try playing around with some way to display some message in the tagbar buffer first. If I can't really come up with anything, then I'll merge this in as is.

raven42:
Updated the code to use the fileinfo so we can track the file size and display a message in the tagbar window in the event the file size limit is exceeded.

alerque (approved):

PR #660, opened 2020-09-23
Add g:tagbar_visibility_symbols

ElPiloto:
This adds `g:tagbar_visiblity_symbols` as a user-facing setting to override the use of `+/#/-` for `public/protected/private` visibility symbols.  If not specified by user, uses pre-existing defaults (`+/#/-`).

### DIFF
--- a/autoload/tagbar/prototypes/basetag.vim
+++ b/autoload/tagbar/prototypes/basetag.vim
@@ -3,6 +3,9 @@ let s:visibility_symbols = {
     \ 'protected' : '#',
     \ 'private'   : '-'
 \ }
+if !empty(g:tagbar_visibility_symbols)
+    let s:visibility_symbols = g:tagbar_visibility_symbols
+endif
 
 function! tagbar#prototypes#basetag#new(name) abort
     let newobj = {}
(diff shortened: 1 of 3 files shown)


ElPiloto:
After implementing this I found this pre-existing pull request: https://github.com/preservim/tagbar/pull/548, which seems to be abandoned.  My implementation should address the requested changes in that conversation.

autoload/tagbar/prototypes/basetag.vim
```diff
@@ -3,6 +3,9 @@ let s:visibility_symbols = {
     \ 'protected' : '#',
     \ 'private'   : '-'
 \ }
```
alerque, line 5: Can this assignment of `s:vilibility_symbols` be dropped entirely? It seems to be a duplicate with the same data being set in the default options array and then parsed into `s:` variables per-option. Whichever place works, I'd rather see the data in one place or the other, not both.

alerque (changes requested):
Thanks for taking the time to contribute. I do agree this implementation looks better than the one in #548. Lets take care of the one caveat about duplicated default option data and I'll be happy to merge this and close the other one.

alerque (approved):

PR #661, opened 2020-09-23
Add g:tagbar_scopestrs

ElPiloto:
Allows mapping scope suffixes (e.g. 'function') to user-defined values.

This adds control over the display of scopes in the Tagbar window (e.g. displaying the `function` suffix as a nice character ⨍ or another string altogether).  If scope is not in `g:tagbar_scopestrs`, no replacement occurs.
```
let g:tagbar_scopestrs = {
\    'class': "\uf0e8",
\    'const': "\uf8ff",
\    'constant': "\uf8ff",
\    'function': "\uf794",
}```


alerque (approved):
Thanks for taking the time to contribute! This LGTM. I'm going to squash-merge since one of the two commits is just a fixup for style issues in the previous one.

PR #663, opened 2020-09-25
Updated instance of a-sync call for tagbar originally implemented by @pangchol

### LINKED ISSUE #532: suggest let s:AutoUpdate function run async
I use tagbar to repacle taglist recently, but I find when I use vim-rtags to jump anohter file, or edist a file, if this file is open first time , the main windows screen will often refreash unnecessary. soon I find it is cause by s:AutoUpdate always jump to tagbrer windows first and then jump back 
 ,  this flow is by in sync and the " autocmd BufEnter " auto call s:AutoUpdate  at the buffer not be completely ready.  so I try using job_start to let s:AutoUpdate Call async and find the refreash disappear. my simple  implementation like this:

function! s:AutoUpdate(fname, force, ...) abort
    let g:tagbar_update_fname = a:fname
    let g:tagbar_update_force = a:force
    let g:tagbar_no_display = a:0 > 0 ? a:1 : 0
    call job_start('sleep 0', {'close_cb':'AutoUpdate_CB'})
endfunc

function! AutoUpdate_CB(channel) abort
    " call tagbar#debug#log('AutoUpdate called [' . fname . ']')
    let fname = g:tagbar_update_fname
    let force = g:tagbar_update_force
    let no_display = g:tagbar_no_display

the AutoUpdate_CB is the old s:AutoUpdate function and just modify the three line ahead.

raven42:
Closes #532 

Update of PR #561 originally implemented by @panchol 

This uses a lambda callback to pass the parameters into the callback routine.

note: I kept the `has('win32')` bypass logic in there as per the comment left by @panchol. I don't have a win32 instance to attempt this on, so I can't speak to the comment about `timer_start()` not working correctly on windows.

Currently this is a partial implementation. It has not been heavily tested.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2655,10 +2655,12 @@ function! s:AutoUpdate(fname, force, ...) abort
     " update the fileinfo
     let no_display = a:0 > 0 ? a:1 : 0
 
-    if !has('win32') && has('lambda')
+    if !has('win32') && has('lambda') && has('timers')
+        call tagbar#debug#log('Performing async call to AutoUpdate_CB')
         call timer_start(0, { -> AutoUpdate_CB(a:fname, a:force, no_display)})
     else
-        call AutoUpdate_CB(a:fname, a:force, a:0 < 0 ? a:1 : '0')
+        call tagbar#debug#log('Performing sync call to AutoUpdate_CB')
+        call AutoUpdate_CB(a:fname, a:force, no_display)
     endif
 endfunc
 
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2649,16 +2649,16 @@ endfunction
 " use timer_start to let tagbar async, this can increase vim open file performance and
 " fix windows blink when open some file.
 function! s:AutoUpdate(fname, force, ...) abort
+    let no_display = a:0 > 0 ? a:1 : 0
+
     if !has('win32') && has('lambda')
-        call timer_start(0, { -> AutoUpdate_CB(a:fname, a:force, a:0 > 0 ? a:1 : '0')})
+        call timer_start(0, { -> AutoUpdate_CB(a:fname, a:force, no_display)})
     else
         call AutoUpdate_CB(a:fname, a:force, a:0 < 0 ? a:1 : '0')
     endif
 endfunc
 
-function! AutoUpdate_CB(fname, force, ...) abort
-    let no_display = a:0 > 0 ? a:1 : 0
-    call tagbar#debug#log('AutoUpdate_CB(' . a:fname . ', ' . a:force . ', ' . a:0 . ')')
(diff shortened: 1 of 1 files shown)


alerque:
I won't be able to give this a real shakedown for a week or so, I'm over my eyeballs. There are a couple other people that have access and might be able to offer a review and any member review approval will work so you can merge this. It looks sane to me I just can't test it right now.

By the way I'd suggest squashing the last 4 commits here into your main one since they are all fixups of that. This should be just the original contribution, a merge commit, and your fixups.

If I forget about this in 1.5 weeks feel free to ping me! Great work recently by the way.

raven42:
So did a little testing with this... as per @noscript's comment on [NERDTree#1170](https://github.com/preservim/nerdtree/issues/1170) where it was stated that the timer starts in the same thread, I added a delay in the timer, then some additional debug logs, then opened a large file to see the behavior. This is indeed correct that the delay will always be present, the `timer_start()` will only delay the operation. So in my testing, when delaying for 5 seconds before starting the `AutoUpdate_CB()` routine, I did see the file load right away and I was able to move the cursor around, however after the 5 second timer, vim did lock up for a bit while the ctags were being processed. So this isn't really a solution for an async call, but just delays the call that is made. A different solution will have to be found.

alerque (changes requested):
As found by @raven42, this doesn't really solve the problem, it only delays it. Marking as Draft so it's obvious we're not ready to merge this until it actually solves the problem.

raven42:
We may want to look at integration with the [async-run](https://github.com/skywind3000/asyncrun.vim) plugin by @skywind3000 to do this. It looks like that may be the best way to do a background process within a single threaded vim instance. It looks like that opens a new buffer in a temp file and all output gets written to that buffer. I'm not sure how the buffer management works in tagbar though, so not sure if this would be feasible.

noscript:
Sorry to invade the thread, but you cannot run vim functions in the background and async-run plugin is not helping there. However what is possible in theory is fire up another vim instance and run some heavy vimL code there and return the results via channel.

noscript:
Using `job_start()` is the best solution in this case but `s:run_system()` will have to be split into two calls. 

Before:

```vim
        silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
```

After:
```vim
        call s:run_system(a:ctags_cmd, function("s:foo"))

        function! s:foo(msg)
            let ctags_output = a:msg
            " do something with ctags_output
        endfunction

" ...

function! s:run_system(cmd, callback) abort
    if exists('g:tagbar_job') && job_status(g:tagbar_job) == "run"
        return
    endif
    let g:tagbar_job = job_start(a:cmd, #{ out_cb: function("s:run_system_post", [a:callback]) })
endfunction

function! s:run_system_post(callback, channel, msg) abort
    let Func = function(a:callback)
    call Func(a:msg)
endfunction
```

PR #666, opened 2020-09-29
Hide signcolumn for tagbar window

wfxr:
`signcolumn` doesn't seem to make sense for tagbar. Hiding it can help users save a little screen space.

**Before:**
image

**After:**
image


alerque (approved):
Thanks for taking the time to contribute @wfxr, this makes a lot of sense to me!

Issue #496, opened 2018-06-30
typescriptreact support


Andilutten:
The tagbar doesnt work with .tsx files. Is there any workarounds ?

```
let g:tagbar_type_typescript = {                                                  
  \ 'ctagsbin' : 'tstags.js', 
  \ 'ctagsargs' : '-f-',                                                           
  \ 'kinds': [                                                                     
    \ 'e:enums:0:1',                                                               
    \ 'f:function:0:1',                                                            
    \ 't:typealias:0:1',                                                           
    \ 'M:Module:0:1',                                                              
    \ 'I:import:0:1',                                                              
    \ 'i:interface:0:1',                                                           
    \ 'C:class:0:1',                                                               
    \ 'm:method:0:1',                                                              
    \ 'p:property:0:1',                                                            
    \ 'v:variable:0:1',                                                            
    \ 'c:const:0:1',                                                              
  \ ],                                                                            
  \ 'sort' : 0                                                                    
\ }  
```

tehdb:
any news on that?

alerque:
@tehdb, @Andilutten, or anybody — please feel free to contribute this as a PR.

(See issue 549 if you are wondering about the late action here! Hopefully we can work through some backlog.)

alerque:
Just a note, since this issue was opened both Vim and Neovim have both added official support for React based file types with their own file type monikers: `javascriptreact` and `typescriptreact`. This will be relevant to anybody who tried to implement this and submit a PR because those are the filetypes we should use now.

mrboli:
@alerque what's required to do this?

Is it just defining the type `typescriptreact` and then creating some kinds that would match up with a tag generator?
I'm still a big confused as to how this works, apparently tagbar *doesn't* look at the tags file?

Would love some help, I'd be willing to take a crack at this.

tsonge:
I'm facing the same issue. I followed the instruction on https://github.com/majutsushi/tagbar/wiki#typescript and it works fine  with .ts files, just not with .tsx files, even though running `tstags myfile.tsx` manually produces the appropriate `tags` file. 

Following the instructions in `:h tagbar-issues` for debugging, it seems like there is no `tagbardebug.log.ctags_out` file produced, unlike in the plain `.ts` case.

joshfrench:
I'm taking a stab at this and running into an issue with integrating `tstags`, which seems to have much more accurate output than `uctags`.

I've added the following to `autoload/tagbar.vim`:
```
[...]

PR #669, opened 2020-10-01
Add typescript mappings for universal ctags

joshfrench:
Re: https://github.com/preservim/tagbar/issues/496

Adds basic Typescript mappings. I ended up skipping `tstags` support, as evidently it doesn't output scopes or enumerations in a compatible format:

```ts
// test.ts
enum Test {
  FOO,
}

class Bar {
  baz() {}
}
```
```sh
$ ctags -f - test.ts
Bar     test.ts      /^class Bar {$/;"       c
FOO     test.ts      /^  FOO,$/;"            e       enum:Test
Test    test.ts      /^enum Test {$/;"       g
baz     test.ts      /^  baz() {}$/;"        m       class:Bar

$ tstags -f - test.ts
Test     test.ts       /^enum Test {$/;"       e       line:1
Bar      test.ts       /^class Bar {$/;"       C       line:5
Bar#baz  test.ts       /^  baz() {}$/;"        m       line:6
```

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -957,6 +957,35 @@ function! tagbar#types#uctags#init(supported_types) abort
         \ {'short' : 'p', 'long' : 'procedures', 'fold' : 0, 'stl' : 1}
     \ ]
     let types.tcl = type_tcl
+    " TypeScript {{{1
+    let type_ts = tagbar#prototypes#typeinfo#new()
+    let type_ts.ctagstype = 'typescript'
+    let type_ts.kinds = [
+        \ {'short' : 'n', 'long' : 'namespaces',    'fold' : 0, 'stl' : 1},
+        \ {'short' : 'i', 'long' : 'interfaces',    'fold' : 0, 'stl' : 1},
+        \ {'short' : 'g', 'long' : 'enums',         'fold' : 0, 'stl' : 1},
+        \ {'short' : 'e', 'long' : 'enumerators',   'fold' : 0, 'stl' : 1},
+        \ {'short' : 'c', 'long' : 'classes',       'fold' : 0, 'stl' : 1},
+        \ {'short' : 'C', 'long' : 'constants',     'fold' : 0, 'stl' : 1},
+        \ {'short' : 'f', 'long' : 'functions',     'fold' : 0, 'stl' : 1},
+        \ {'short' : 'p', 'long' : 'properties',    'fold' : 0, 'stl' : 1},
+        \ {'short' : 'v', 'long' : 'variables',     'fold' : 0, 'stl' : 1},
+        \ {'short' : 'm', 'long' : 'methods',       'fold' : 0, 'stl' : 1},
+    \ ]
+    let type_ts.sro        = '.'
+    let type_ts.kind2scope = {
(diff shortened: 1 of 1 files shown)


autoload/tagbar/types/uctags.vim
```diff
@@ -957,6 +957,35 @@ function! tagbar#types#uctags#init(supported_types) abort
...
     \ ]
     let types.tcl = type_tcl
+    " TypeScript {{{1
+    let type_ts = tagbar#prototypes#typeinfo#new()
+    let type_ts.ctagstype = 'typescript'
+    let type_ts.kinds = [
+        \ {'short' : 'n', 'long' : 'namespaces',    'fold' : 0, 'stl' : 1},
+        \ {'short' : 'i', 'long' : 'interfaces',    'fold' : 0, 'stl' : 1},
+        \ {'short' : 'g', 'long' : 'enums',         'fold' : 0, 'stl' : 1},
+        \ {'short' : 'e', 'long' : 'enumerators',   'fold' : 0, 'stl' : 1},
```
raven42, line 967: Does it need both `enums` and `enumerators`? Does ctags output two different types?

raven42:
Does [ctags.vim](https://github.com/preservim/tagbar/blob/master/autoload/tagbar/types/ctags.vim) need to be updated as well? Do you have a version of Exuberant Ctags you can test with?

alerque (dismissed):
Thanks for taking the time to contribute this. I know it saves a lot of people time when their file types work out of the box rather than having to track down how to configure it.

LGTM, maybe deal with @raven42's comment but when you guys are ready go for it.

autoload/tagbar/types/uctags.vim
joshfrench: Ah, that should read `enumerations` -- the actual members of an enum. From the sample file above:

```
enum Test {
    FOO
}

Test    test.ts      /^enum Test {$/;"       g
FOO     test.ts      /^  FOO,$/;"            e       enum:Test
```

alerque:
> Does [ctags.vim](https://github.com/preservim/tagbar/blob/master/autoload/tagbar/types/ctags.vim?rgh-link-date=2020-10-01T20%3A48%3A05Z) need to be updated as well? Do you have a version of Exuberant Ctags you can test with?

I doubt Exuberant supports it, TypeScript came out after it stagnated. And frankly I don't care. I'm still not at the point of deliberately ripping out Exuberant support, but I'm close and think that no time should be wasted adding new support for it.

raven42:
> > Does [ctags.vim](https://github.com/preservim/tagbar/blob/master/autoload/tagbar/types/ctags.vim?rgh-link-date=2020-10-01T20%3A48%3A05Z) need to be updated as well? Do you have a version of Exuberant Ctags you can test with?
> 
> I doubt Exuberant supports it, TypeScript came out after it stagnated. And frankly I don't care. I'm still not at the point of deliberately ripping out Exuberant support, but I'm close and think that no time should be wasted adding new support for it.

Ok sounds good to me. We can skip exuberant tags then.

raven42 (approved):

alerque (approved):

joshfrench:
Thanks y'all!

PR #674, opened 2020-10-16
Allow more granular control of `tagbar_compact`

### LINKED ISSUE #673: More granular `g:tagbar_compact`
Currently, setting `g:tagbar_compact = 1` will disable the short help message at the top and remove spaces between top-level scopes. Is there a way to only disable the help message and keep the spacing between scopes?

I'm happy to work on this as a feature if there's no other way to do it in the plugin currently.

maujim:
Closes #673

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1952,7 +1952,7 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
             for tag in curtags
                 call s:PrintTag(tag, 0, output, a:fileinfo, a:typeinfo)
 
-                if !g:tagbar_compact
+                if g:tagbar_compact != 1
                     call add(output, '')
                 endif
             endfor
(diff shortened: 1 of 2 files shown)


autoload/tagbar.vim
```diff
@@ -1952,7 +1952,7 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
             for tag in curtags
                 call s:PrintTag(tag, 0, output, a:fileinfo, a:typeinfo)
 
-                if !g:tagbar_compact
+                if g:tagbar_compact != 1
```
alerque, line 1955: Do we need to test here for the case where the variable is not set at all? Is being set to *something* a guarantee at this point due to the way we load preferences or might this not exist at all?

maujim: No tests needed. All the variables get initialized to their defaults in `tagbar/plugin/tagbar.vim`.

alerque (approved):
Thanks for taking the time to contribute @maujim! LGTM.

PR #675, opened 2020-10-17
sq: would mess if set key mapping to ''

Shane-XB-Qian:
for now, user is not able to disable some key mapping in tagbar window, 
or e.g: `let g:tagbar_map_togglesort = ''` but current code can not handle this, 
'mapping' would be messed, would map to unexpected result since `key` actually was 'empty'.

// fix it and made user be able to disable key mapping.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -511,8 +511,10 @@ function! s:MapKeys() abort
             let keys = def
         endif
         for key in keys
-            execute 'nnoremap' . map_options . key .
-                        \ ' :call <SID>' . func . '<CR>'
+            if !empty(key)
+                execute 'nnoremap' . map_options . key .
+                            \ ' :call <SID>' . func . '<CR>'
+            endif
         endfor
         unlet def
     endfor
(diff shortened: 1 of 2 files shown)


raven42 (dismissed):
Looks good to me

doc/tagbar.txt
```diff
@@ -411,11 +411,15 @@ q             Close the Tagbar window.
...
 These mappings can be redefined with the given map options. The argument can
 be either a string or a |List| of strings. In the latter case the
-functionality will be assigned to all of the keys in the list. For example, if
-you want to remap the sort toggling functionality to "r":
+functionality will be assigned to all of the keys in the list.
+For example, if you want to remap the sort toggling functionality to "r":
 >
         let g:tagbar_map_togglesort = "r"
 <
+For example, or if you want to disable this mapping, then set to '':
```
alerque, line 419:
```suggestion
Alternatively, if you want to disable this mapping, then set to '':
```

autoload/tagbar.vim
```diff
@@ -2061,34 +2063,33 @@ function! s:PrintHelp() abort
...
         silent  put ='\" --------- General ---------'
-        silent  put ='\" ' . s:get_map_str('jump') . ': Jump to tag definition'
-        silent  put ='\" ' . s:get_map_str('preview') . ': As above, but stay in'
-        silent  put ='\"    Tagbar window'
-        silent  put ='\" ' . s:get_map_str('previewwin') . ': Show tag in preview window'
-        silent  put ='\" ' . s:get_map_str('nexttag') . ': Go to next top-level tag'
-        silent  put ='\" ' . s:get_map_str('prevtag') . ': Go to previous top-level tag'
-        silent  put ='\" ' . s:get_map_str('showproto') . ': Display tag prototype'
-        silent  put ='\" ' . s:get_map_str('hidenonpublic') . ': Hide non-public tags'
+        if !empty(s:get_map_str('jump')) | silent  put ='\" ' . s:get_map_str('jump') . ': Jump to tag definition' | endif
```
alerque, line 2066: This and the following ~28 lines are 95% repeated boiler plate. Perhaps we could abstract this into a function and just iterate over a list doing the same gymnastics for each one with only 5% of the code. Not a blocker if you don't have time to refactor this, but it would sure make this a lot prettier.

alerque (changes requested):
Looks great, thanks for the contribution! I do have one minor quip with the documentation that should be an easy fix. Also one possible refactoring that would make this all a lot prettier if you have time, but it can go through without too.

autoload/tagbar.vim
Shane-XB-Qian: perhaps tidy up those into a func may make it looks clean, but keyword & description each of them was different, perhaps would not reduce items/letters too much, specially to me, i was trying to reduce as less modification as possible to make it less side-effect. so perhaps just let it just like that for now, or may refactor this help func again in future with others.

alerque (approved):

autoload/tagbar.vim
alerque: The different keyword and description pairs could be in a dictionary list that was iterated over. The code logic (`if`, `put`, etc.) is all 100% identical and could be abstracted to a function. I've opened #681 to track this if anybody wants to tackle it.

Shane-XB-Qian: keyword &  description were there already before this pr, i did not want to change too much;
compared to author, to me, i think i'd better less modification (not refactor too much) as better,
or author (or you?) to refactor perhaps better, since you knew that should be more clear than me.
anyway, thanks merged this.

Issue #672, opened 2020-10-14
Problems opening perl files


diegok:
Since I've updated around a month ago I'm having lots of errors opening perl files and probably others:

```
Error detected while processing function tagbar#ToggleWindow[2]..<SNR>90_ToggleWindow[9]..<SNR>90_OpenWindow[81]..<SNR>90_AutoU
pdate[63]..<SNR>90_ProcessFile[146]..<SNR>110_sortTags[2]..tagbar#sorting#sort[6]..<SNR>114_compare_by_kind:
line    3:
E716: Key not present in Dictionary: m
E15: Invalid expression: typeinfo.kinddict[a:tag1.fields.kind] <# typeinfo.kinddict[a:tag2.fields.kind]
E716: Key not present in Dictionary: m
```

I had to disable this plugin as it was impossible to work with it and I was updating and re-checking since this problems started. I've seen other issues with similar problems for JS/bash so I was just waiting to see how it goes but I don't see any progress. I'm using universal-ctags and I've seen in those issues that this can be related, but it doesn't exists a package for the old "exuberant-ctags" on my distro...

If there is any more info I can collect to help understand this problem just let me know.

raven42:
Do you have a sample perl script that generates this error? I only have a few small perl scripts and I do not see this error when I open them up.

diegok:
Sure, I've just reproduced this problem here:

- https://github.com/diegok/resque-perl/blob/master/lib/Resque/Job.pm

raven42:
I have a potential fix to address this globally for unknown `kinds`. I am not sure what the specifics of the perl `m` kind is, so #676 will not actually add proper handling for the `m` kind, but only prevent any tagbar failures when an unknown kind is found.

As a followup for this, someone more familiar with perl can update the perl type definition to include the proper `m` kind (and possibly the `a` kind too).

diegok:
Hey @raven42 I can help out with this. I'm not familiar with ctags or vim script, but I'm all in to help improve this.

From what you've said and what I've pasted, `m` looks like `method` as in: `fail	lib/Resque/Job.pm	/^sub fail {$/;"	m	class:Resque::Job` but then I can see an `s` with the same line but without the class at the end, so I'm probably missing something here :-/

On the other hand the `a` looks like `attribute` which is what `has` is used for. To give you some more context, this is nos native perl but an convention provided by several libraries in CPAN which Moose is the most used and the one used by this example.

Just let me know if I can help in any other way and thank you!.

[...]

PR #677, opened 2020-10-20
Add missing attributes and methods kinds on perl type #672

diegok:
This adds info on two missing kinds over the perl type as explained by @raven42 on https://github.com/preservim/tagbar/pull/676

### DIFF
--- a/autoload/tagbar/types/ctags.vim
+++ b/autoload/tagbar/types/ctags.vim
@@ -466,9 +466,11 @@ function! tagbar#types#ctags#init(supported_types) abort
     let type_perl.kinds     = [
         \ {'short' : 'p', 'long' : 'packages',    'fold' : 1, 'stl' : 0},
         \ {'short' : 'c', 'long' : 'constants',   'fold' : 0, 'stl' : 0},
+        \ {'short' : 'a', 'long' : 'attributes',  'fold' : 0, 'stl' : 0},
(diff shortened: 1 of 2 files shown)


alerque (commented):
Thanks for taking the time to contribute @diegok! I hope this will benefit other Perl users.

Did you actually check that these types are output by both the legacy Exuberant `ctags` and the modern Universal `ctags`? If not maybe we should just add these to *uctags*. We're phasing out Exuberant support anyway but I'm not sure there is a good reason to add something it doesn't actually support.

diegok:
No. I've not checked exuberant-ctags. I only have universal. I've added to it just because I've seen both were the same. Do you need me to remove those changes it from the legacy one?

diegok:
I've removed the changes on the legacy exuberant-ctags file :+1:

alerque (approved):
I think this looks good now. There *might* not have been any harm in specifying types that e-ctags doesn't output, but I wasn't sure on that so I'd rather not touch it except for confirmed working changes. Unless @raven42 has other ideas I say this is good to go.

raven42:
Looks good to me. Thanks for the PR @diegok. I've merged this one in as well as #676 with the extra protection around the unknown kinds.

diegok:
Tank you @raven42 for pointing me on the right direction!

PR #683, opened 2020-10-26
Scoped Highlighting

### LINKED ISSUE #376: Current function not identified if it contains a #define
This simple code can reproduce the issue on 01f57ac5 (vim 7.4, ctags 5.9~svn20110310,):

```
#define WORKING 1

int main() {
#define NOTWORKING 2

    return 1;
}
```
Place the cursor on the `return` statement and open tag bar. It show the current context as in `NOTWORKING` macro, instead of the `main` function.

See this example:

Screenshot of problem reproduced live

While if the cursor is above the macro, it's working as expected:

Screenshot of problem not reproduced with caret above the macro

raven42:
Closes #376

Attempt to change the scoped highlighting of the tagbar window. This will now look only for 'stl' tag types when looking for the nearby tag when looking to highlight. It does however take into account if the cursor is on the same line as the tag and highlights it then as well.

Add debug function tagbar#printfileinfo() that can be used to print debug info to the tagbar debug file.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1766,10 +1766,7 @@ function! s:create_pseudotag(name, parent, kind, typeinfo, fileinfo) abort
                     \ pseudotag.path . a:typeinfo.sro . pseudotag.name
     endif
     let pseudotag.depth = len(split(pseudotag.path, '\V' . escape(a:typeinfo.sro, '\')))
-    call tagbar#debug#log('pseudotag [' . pseudotag.strfmt() . '] depth set to ' . pseudotag.depth)
-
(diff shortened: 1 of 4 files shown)


alerque (approved):
It might be worth rebasing to squash the "remove debug" commits together for the sake of people updating plugins and actually reading the changelog, but otherwise this looks good.

raven42:
> It might be worth rebasing to squash the "remove debug" commits together for the sake of people updating plugins and actually reading the changelog, but otherwise this looks good.

I will do a squash and merge so there is only a single commit that gets merged in with all the changes.

PR #684, opened 2020-10-27
Add option to print the tag linenumber in the tagbar window

### LINKED ISSUE #490: Line number of the Tag
**TL;DR;** Can we show the line number of the tag in the Tag Window?

Say if the tag is on line 131 of the file, that way I may navigate to a tag by `:131` without shifting the focus to the Tag Window.

Long Story:

I have the following settings in my `~/.vimrc`:
```
let g:tagbar_compact = 1
let g:tagbar_show_linenumbers = 1
let g:tagbar_type_javascript = {
  \ 'ctagsbin' : 'ctags',
  \ 'ctagsargs' : '-f - --fields=+n',
  \ }
```
When I am working on an `.js` file, from `set ft?` I can see `filetype=javascript` and the functions (though being shown as "namespace" and all the `if` and `function` are there..), but not the line numbers.

Then I figured out that the "line numbers in Tag Window" refers to .. the tag window itself, not the tags.
Is there a setting which we can see something like `(line ##) function_name : function` ?

raven42:
Closes #490

Add `g:tagbar_print_lnum` option to print the line number of the tag.


alerque (dismissed):

alerque (approved):

Shane-XB-Qian:
@raven42  is it possible reduce the hint chars from '[line 123]' to '[l#123]' or something else, for saving space..

raven42:
We could certainly look at that. Is this something you want to use? If so we can re-open this issue and add some more options.

Shane-XB-Qian:
well, i think we did not need a 'option' or it had 'g:tagbar_show_tag_linenumbers' already, 
i think short/compact '[123]' should be good enough.. if user opened set 'g:tagbar_show_tag_linenumbers', he/she should be awared that is a linenum already.
// i can push a pr for this i you agreed this, and looks the pattern `[0-9]*` for hl was not accurate.

raven42:
Sure a PR would be welcome.

Is the hl pattern not working correctly? It was in all my testing.

Shane-XB-Qian:
pr #687 opened for this.

PR #685, opened 2020-10-27
Add g:tagbar_show_tag_count option

### LINKED ISSUE #290: Possible to show count of functions/variables/etc ?
I love this plugin!

Is it possible to configure it to show a count of the number of items in each group? So, assuming everything is folded, rather than:

---

```
> macros

> variables

> functions
```

---

... it would instead show:

---

```
> macros (4)

> variables (17)

> functions (125)
```

---

raven42:
Closes #290

This option will show the tag kind count next to the kind label in the tagbar window. Note: this is limited to only the actual tag kinds. It does not list the count of members for a scoped value such as union or class.

Example:
```
" Press <F1>, ? for help

> macros (3)

> variables (1)
    global_union

> functions (6)
    function5(int i)
    funtion1(int i)
    funtion2(int i)
    funtion3(int i)
    funtion4(int i)
    main(int argc,char * argv[])
```

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1967,8 +1967,8 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
             endif
 
             let padding = g:tagbar_show_visibility ? ' ' : ''
-            let tag_count = ' (' . len(curtags) . ')'
             if g:tagbar_show_tag_count
+                let tag_count = ' (' . len(curtags) . ')'
                 call add(output, foldmarker . padding . kind.long . tag_count)
             else
                 call add(output, foldmarker . padding . kind.long)
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1967,7 +1967,12 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
             endif
 
             let padding = g:tagbar_show_visibility ? ' ' : ''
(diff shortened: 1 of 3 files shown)


autoload/tagbar.vim
```diff
@@ -1967,7 +1967,12 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
             endif
 
             let padding = g:tagbar_show_visibility ? ' ' : ''
-            call add(output, foldmarker . padding . kind.long)
+            let tag_count = ' (' . len(curtags) . ')'
+            if g:tagbar_show_tag_count
```
alerque, line 1971:
```suggestion
            if g:tagbar_show_tag_count
                let tag_count = ' (' . len(curtags) . ')'
```

As long as we have the feature behind a flag we might as well save the processing time generating the string if we're not going to use it.

alerque (changes requested):

autoload/tagbar.vim
raven42: fixed

raven42:
Regarding this feature, should it just be visible all the time? or should we keep it behind a configuration flag?

alerque:
I would prefer the un-clutterd look of off by default. I don't mind the options being there for people that want them, but I'm kind of a minimalist.

raven42:
Ok sounds good to me.

alerque (approved):

PR #686, opened 2020-10-27
Help visibility

### LINKED ISSUE #410: request: add option to always show expanded help message:
normally you have to press `F1` or `?` to see the help message. i was wondering if there could be an option for it, since on larger monitors there's usually space left over that may as well show it.

raven42:
Closes #410

Add `g:tagbar_help_visibility` option to always show the help dialog

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1998,7 +1998,7 @@ function! s:PrintKinds(typeinfo, fileinfo) abort
     endfor
 
     let outstr = join(output, "\n")
-    if g:tagbar_compact && s:short_help
+    if g:tagbar_compact && !g:tagbar_help_visibility && s:short_help
         silent 0put =outstr
     else
         silent  put =outstr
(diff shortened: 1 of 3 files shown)


alerque (approved):

Shane-XB-Qian:
@raven42 perhaps better if adding another option for 'let g:tagbar_compact = `3`' :
  0: Show short help and blank lines between top-level scopes
  1: Don't show the short help or the blank lines.
  2: Don't show the short help but show the blank lines.
  **3: Show the short help but don't  show the blank lines.**
probably just needed a short help hint -vs- full version help items always showed there.

raven42:
@Shane-XB-Qian we could add that, but that would be a separate feature request. This feature request is about having the full help dialog visible all the time. I'm mainly trying to go through a lot of the older issues and attempt to close some out that are just quick little options like this so we can get the issue count down without a huge amount of rework. If this is something you'd like to have, then we can certainly look at a pull request for it.

Shane-XB-Qian:
well, i just thought that would be more make sense, usually tagbar showed at corner window, its space should be very limited, so generally short help would be more helpful, since there only some key-mappings (-vs- no others info can get there), mostly should be no needed to show all the time. 'let g:tagbar_compact = 3' should be more useful than 'g:tagbar_help_visibility'. up to you, i am just saying my point.

Shane-XB-Qian:
yes, i agree, for now, such 'g:tagbar_help_visibility' perhaps not very useful, but that's a historical background/issue..
so let it be.. :)

PR #687, opened 2020-10-28
shane: compact tag line# should be good enough if user set it

Shane-XB-Qian:
pr for #684
short/compact '[123]' should be good enough.. if user opened set 'g:tagbar_show_tag_linenumbers', he/she should be awared that is a linenum already.
// the pattern [0-9]* for hl was not accurate, line# should be always existed or there was no that tag existed.
@raven42  pls review, and merge if ok.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2202,7 +2202,7 @@ function! s:HighlightTag(openfolds, ...) abort
         " If printing the line number of the tag to the left, and the tag is
         " visible (I.E. parent isn't folded)
         if g:tagbar_show_tag_linenumbers == 2 && tagline == tag.tline
-            let pattern = '/^\%' . tagline . 'l\s*' . foldpat . '[-+# ]\[line [0-9]*\] \?\zs[^( ]\+\ze/'
+            let pattern = '/^\%' . tagline . 'l\s*' . foldpat . '[-+# ]\[[0-9]\+\] \?\zs[^( ]\+\ze/'
         else
             let pattern = '/^\%' . tagline . 'l\s*' . foldpat . '[-+# ]\?\zs[^( ]\+\ze/'
         endif
--- a/autoload/tagbar/prototypes/normaltag.vim
+++ b/autoload/tagbar/prototypes/normaltag.vim
@@ -37,9 +37,9 @@ function! s:strfmt() abort dict
     let prefix = self._getPrefix()
 
     if g:tagbar_show_tag_linenumbers == 1
-        let suffix .= ' [line ' . self.fields.line . ']'
+        let suffix .= ' [' . self.fields.line . ']'
     elseif g:tagbar_show_tag_linenumbers == 2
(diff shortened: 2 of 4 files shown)


raven42 (commented):
The original feature request was to use the syntax `(line #)` so I was trying to make sure we follow that request. I used `[` and `]` characters instead of parenthesis because the highlight pattern matching used parenthesis to determine highlighting the tag vs function parameters and so that changed the color of the tag if the line numbers were on the left.

If we want to allow a short version, then I feel that should be done through another possible setting for the `g:tagbar_show_tag_linenumbers` field. Maybe something like a `g:tagbar_show_tag_linenumbers = 3` to show shorter line number syntax to the right of the tag, and `g:tagbar_show_tag_linenumbers = 4` to show the shorter line number syntax to the left of the tag. However I'm not going to be using this feature in my own setup, so if there is a stronger consensus to only have the two options and stick with the shorter variant, then I'm fine with this.

Shane-XB-Qian:
@raven42 just like we talked at #686, we'd better just short version and/or that should be good enough for this at least.
otherwise, would become another historical issue again .. -_-#  we did not need so many options.. :)

alerque:
I think I'm okay with a short syntax by default (since this is off-by-default people enabling it will understand what they are getting, but honestly any use of `#` and even the brackets is feeling cluttered to me. Have we tried just an upper cases `L` prefix like `L14`? This would be more in line with GitHub's and many other code viewing UIs and links, I think it might be a more familiar notation than any parenthesis, brackets, or hashes. I realize a lot of code debuggers also use lowercase `l` for this, but they often do so with a colon as well `l:14`.

Shane-XB-Qian:
`L123` looks ok, but put tag name e.g `SqFooFunc L123` or `L123 SqFooFunc`, that perhaps looks a bit confused.
so far, looks `[123]` is better choice.

raven42:
Ok sure. I'm fine with any approach. This looks ok to me.

raven42 (approved):

Shane-XB-Qian:
@raven42  i means if `g:tagbar_show_tag_linenumbers` had 3 or 4 etc, then `g:tagbar_compact` probably should have 3 too.. or let it be all...

@alerque  `L123` can be a func/var too in most coding langs, but `[123]` mostly can not, so perhaps `[123]` better.

Shane-XB-Qian:
as for the pattern `[0-9]*` in the code, looks it should be `[0-9]\+`. ok, this is all about this pr, for compact showing line# and this pattern correction.

raven42:
merged in... thanks for the contribution

Shane-XB-Qian:
thx for merged. 
though there was perhaps a bit flaw actually, if set this on and line# showed at right, looks it took same 'syn highlight' like 'parameter' or 'return type' etc such strings of ori func/tag; but this is regardless compact or not, not an issue, just maybe would not look very nice/perfect.

Shane-XB-Qian:
> not an issue, just maybe would not look very nice

pr#701 opened for this.

PR #688, opened 2020-10-28
Add `g:tagbar_no_autocmds` option to disable almost all autocmds

### LINKED ISSUE #578: Disable auto refresh/update cursor
How to avoid `updatetime` using by tagbar? I don't want to do a `refresh` when using tagbar, I want it just echo tags and never update/follow up cursor.

raven42:
Closes #578 

Add `g:tagbar no autocmds` option to disable almost all autocmds.

The only autocmds left will be the QuitPre event and the WinEnter event to handle when tagbar is the last window to make sure it gets closed out correctly.


alerque (approved):

raven42:
Waiting for response from feature requester to see is this is what they are looking for.

raven42:
Going to deliver this one as is since we have not heard back from the issue requester. If we do hear that this isn't the desired behavior we can re-open the issue and resolve it then.

quantum-booty:
This option currently disable highlight updates, but is it possible to update the highlight for the current tag the moment I move my cursor to the tagbar? That way I can see my current position while I'm in the tagbar.

alerque (approved):

raven42:
> This option currently disables all highlight updates, but is it possible to update the highlight for the current tag the moment I move my cursor to the tagbar (And clear the highlight after I move out)? That way I can see my current position while I'm in the tagbar.

@quantum-booty I do not know of a clean way to do this. The tagbar highlighting works by examining the current active buffer and looking for the nearest tag to the cursor. The purpose of this PR was to disable any processing with autocmds so as not to use up processing time during normal usage. Without doing any processing, we wouldn't be able to determine the nearest tag when the cursor moves. The main delay in processing for the actual issue request for this PR is, as I understand it, the delay being caused by the actual tag searching. Once we know what tag to highlight, that operation to actually jump to and highlight the tag is minimal. So the purpose of this PR is to prevent any processing for the actual nearby tag searching.

Without the ability to search the current tag in real time, the only option I can think of would be to trigger on entry of the tagbar window, then immediately jump back to the file window, grab the cursor position and search for the nearby tag, then switch back to the tagbar window again and jump to and highlight the tag. However to further complicate things I'm not sure of a way to tell how to find the correct window. Say for example you have two open files in split windows, and then three other splits with different plugins (say tagbar, nerdtree, and undo-tree for example). So now you have a total of 5 window splits, and only two of them are files. But say the last window you were in was nerdtree, then you switch to tagbar. How would we tell which window to switch back to in order to determine the cursor position.

The request you are making I believe falls outside the intent of this PR and would take a lot of coding effort to handle these end cases. I'm not sure of a clean way to do it to handle the end cases without any major rework. If you are able to come up with something, I'd be happy to look at a PR and help integrate it if needed. I'm just not sure how best to go about that at present.

I'll merge this one in for now and if you want to follow it up with another PR @quantum-booty, we'd be happy to look at a better solution.

PR #692, opened 2020-10-29
Add g:tagbar_scroll_offset configuration

### LINKED ISSUE #564: Add option for tagbar tag position in the window
Add option to control where the current tag is positioned in the tagbar window.
It would allow, for example, to always keep the current tag at the center/top/bottom of the tagbar window.
This option is similar to the one proposed [here](https://github.com/majutsushi/tagbar/issues/504), but applies to the tagbar window.

raven42:
Closes #564

Add option for `g:tagbar_scroll_offset` to issue a `setlocal scrolloff=#` during tagbar window init

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -947,8 +947,8 @@ function! s:InitWindow(autoclose) abort
     setlocal nomodifiable
     setlocal textwidth=0
 
-    if g:tagbar_scroll_offset > 0
-        execute 'setlocal scrolloff=' . g:tagbar_scroll_offset
+    if g:tagbar_scrolloff > 0
+        execute 'setlocal scrolloff=' . g:tagbar_scrolloff
     endif
 
     if g:tagbar_show_balloon == 1 && has('balloon_eval')
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -947,6 +947,10 @@ function! s:InitWindow(autoclose) abort
     setlocal nomodifiable
     setlocal textwidth=0
 
+    if g:tagbar_scroll_offset > 0
+        execute 'setlocal scrolloff=' . g:tagbar_scroll_offset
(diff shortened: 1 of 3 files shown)


autoload/tagbar.vim
```diff
@@ -947,6 +947,10 @@ function! s:InitWindow(autoclose) abort
     setlocal nomodifiable
     setlocal textwidth=0
 
+    if g:tagbar_scroll_offset > 0
```
alerque, line 950: Can we call this `g:tagbar_scrolloff` to match the vim setting name that actually implements it?

alerque (changes requested):

autoload/tagbar.vim
raven42: done

alerque (approved):

plugin/tagbar.vim
```diff
@@ -101,6 +101,7 @@ function! s:setup_options() abort
         \ ['position', default_pos],
         \ ['previewwin_pos', previewwin_pos],
         \ ['scopestrs', {}],
+        \ ['scrolloff', 0],
```
alerque, line 104: I don't use scrolloff much myself (except when I set it to 999 to keey the current line centered for some operations) so I don't know if it would make any sense, but is there any logic in setting the default to the parent buffer's setting of `scrolloff`?

raven42: that's a good idea.... done

raven42: Hmm... actually on second thought we shouldn't need to do this.

By setting a default of `0`, then we do not even issue the `setlocal scrolloff=#` so the global `scrolloff` value will be honored.

alerque: In that case that's better, only set this if you don't want the inherited value. :+1:

alerque (approved):

PR #694, opened 2020-10-29
Fix issue with tagbar changing the global scrolloff value.

### LINKED ISSUE #693: tagbar is changing global scrolloff value
In RenderContent() tagbar is changing the global scrolloff value. This should be using setlocal instead to only adjust the value for the tagbar window.

raven42:
Closes #693

Tagbar is overwriting the global scrolloff value which can affect other windows. Change the behavior so it does a setlocal instead when needed so this only impacts the tagbar window

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1912,7 +1912,7 @@ function! s:RenderContent(...) abort
         normal! zt
         call cursor(saveline, savecol)
 
-        execute 'setlocal scrolloff=' . scrolloff_save
+        let &l:scrolloff = scrolloff_save
     else
         " Make sure as much of the Tagbar content as possible is shown in the
         " window by jumping to the top after drawing
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1906,13 +1906,13 @@ function! s:RenderContent(...) abort
     if !empty(tagbar#state#get_current_file(0)) &&
      \ fileinfo.fpath ==# tagbar#state#get_current_file(0).fpath
         let scrolloff_save = &scrolloff
-        set scrolloff=0
+        setlocal scrolloff=0
 
         call cursor(topline, 1)
         normal! zt
         call cursor(saveline, savecol)
 
-        let &scrolloff = scrolloff_save
+        execute 'setlocal scrolloff=' . scrolloff_save
     else
         " Make sure as much of the Tagbar content as possible is shown in the
         " window by jumping to the top after drawing


autoload/tagbar.vim
```diff
@@ -1906,13 +1906,13 @@ function! s:RenderContent(...) abort
...
         let scrolloff_save = &scrolloff
-        set scrolloff=0
+        setlocal scrolloff=0
 
         call cursor(topline, 1)
         normal! zt
         call cursor(saveline, savecol)
 
-        let &scrolloff = scrolloff_save
+        execute 'setlocal scrolloff=' . scrolloff_save
```
alerque, line 1915:
```suggestion
        let &l:scrolloff = scrolloff_save
```
I see why execute gets the job don, but I'm not a fan of composing commands as strings and passing them to be evaled when there is a direct method available.

alerque (changes requested):

alerque (approved):

Shane-XB-Qian:
looks this change did not work well, e.g `set scrolloff=0` at line#2126
did not dig code detail, just chk looks it seems like that .. and influenced others windows if set value to this option.

raven42:
@Shane-XB-Qian can you please elaborate? This change was to update tagbar to use `setlocal` instead of `set` so it is not changing a global setting.

Shane-XB-Qian:
sorry, i have no time to dig detail, but for now obviously not correct:
you can e.g start vim with `let g:tagbar_scrolloff=16` and `set so=10`, you can see globally value had been influenced.

raven42:
@Shane-XB-Qian I think you must have not pulled in this fix. This fix specifically fixes the issue you are referring to. The `g:tagbar_scrolloff` field was introduced in #692. But this fix was to correct the behavior which already existed in the tagbar code where it was setting the global `scrolloff`. Please make sure you pull in the latest and try again. I tested with the values you specified and when I issue a `echo &scrolloff` in the tagbar window, I see 16, and then without changing anything if I issue that in another window, I see a value of 10.

alerque:
> What this _let &l:scrolloff = scrolloff_save_ for?

That sets the *buffer local* scroll off value to the previously saved value.

> And check line#2126 there is _set scrolloff=0_ too.

This is a second instance of the problem we were fixing. This PR fixed one instance, that is a different one.

> I felt those code original author was trying to save _&so_ value, then _zt_, then recover.

The original code was saving a value, setting it to something else, doing the action, then restoring the value. The problem was the set it was doing was a global set instead of a local one, then restored it also by doing a global set instead of local.

@raven42 The instance you fixed is in `s:RenderContent()`, the one tripping up @Shane-XB-Qian is another instance of the same kind of error in `s:RenderKeepView()` that we didn't touch at all.

raven42:
Honestly I'm not sure the purpose of this little block of code. I found this while looking at another issue and thought I'd fix it so we aren't changing the global scrolloff value.

raven42:
I've pushed up another PR (#697) that will address the other instance of this bug. I've also looked for any other instances where the scrolloff is being adjusted, and this looks like the only other place.

Issue #516, opened 2018-12-05
Tags returning function even when outside the scope


kkanungo17:
I use tagbar with lightline to show the function the cursor currently is in, which works fine, but when the cursor goes out of the scope but before going into another function, the previous function still stays.

This is correct:
![screen shot 2018-12-05 at 19 21 44](https://user-images.githubusercontent.com/11506039/49517828-15a13e00-f8c3-11e8-889a-f097cbb5dc6f.png)

This is not:
![screen shot 2018-12-05 at 19 21 52](https://user-images.githubusercontent.com/11506039/49517842-22be2d00-f8c3-11e8-99e7-4ecd9976d25d.png)

bam80:
duplicate: #524 (also valid for Bash scripts)

Any progress here? Very  annoying..

bam80:
@kkanungo17 FYI: you don't need any additional plugin to show the function in statusline.
I use following:
```
:let &statusline = '%f %([%{tagbar#currenttag(''%s'','''')}] %)%h%m%r%=%-14.(%l,%c%V%) %P'

kkanungo17:
Thanks for the FYI, but I use lightline for a couple of other stuff and eye-candy as well. This issue is still annoying though.

bam80:
@majutsushi any comments?

bam80:
@kkanungo17, meanwhile, you are welcome to try https://github.com/dr-chip-vim-scripts/StlShowFunc.

It doesn't need neither tags nor any dependency plugins for it's functioning, does exactly the status line manipulation to show the name of the current function, and handles the scope correctly.

alerque:
@bam80 and @kkanungo17 I would be happy to review a contribution if anybody wants to submit a PR that fixes this. I'm not entirely convinced it is a _bug_ so much as something you'd like configured differently, but if you can figure out how to make it work as expected and introduce that as on option I think that would be great. Keep in mind that in many languages there isn't an explicit closure at all, only opening a new heading/function/scope will close the previous one, so this might need some finesse not to break all languages.

bam80:
> in many languages there isn't an explicit closure at all

In those cases there are indenting or such, am I missing something?
IMO it's clearly the bug since the feature is to show where you are and not what's the upper function name.

alerque:
[...]

PR #696, opened 2020-10-30
Scoped kinds

### LINKED ISSUE #508: Use tag end field to not show as being part of previous tag
When using tags in large files with many functions there are some times when the tagbar status line will show the tag above the cursor, even if outside that tags' scope:

OK:
image

Also OK:
image

Not OK:
image

I have looked at turning on all of the tag fields for cpp and this does indeed generate an end tag, but it is ignored by tagbar:
image

The tags file was built with ```ctags --fields=* test.cpp```

If you require more information please let me know

raven42:
Closes #508, #516

Add --fields=e (end) to the ctags field types to look for the end of the scope.
Update the `s:GetNearbyTag()` routine to use this scope to look for the correct tag.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2158,9 +2158,9 @@ function! s:HighlightTag(openfolds, ...) abort
     let force = a:0 > 0 ? a:1 : 0
 
     if a:0 > 1
-        let tag = s:GetNearbyTag('highlight', 0, a:2)
+        let tag = s:GetNearbyTag('nearest-stl', 0, a:2)
     else
-        let tag = s:GetNearbyTag('highlight', 0)
+        let tag = s:GetNearbyTag('nearest-stl', 0)
     endif
     if !empty(tag)
         let tagline = tag.tline
@@ -2620,7 +2620,7 @@ function! s:OpenParents(...) abort
     if a:0 == 1
         let tag = a:1
     else
-        let tag = s:GetNearbyTag('parent', 0)
+        let tag = s:GetNearbyTag('nearest', 0)
     endif
 
     if !empty(tag)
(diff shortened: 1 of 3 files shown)


autoload/tagbar.vim
```diff
@@ -1525,6 +1525,22 @@ function! s:ProcessTag(name, filename, pattern, fields, is_split, typeinfo, file
         let taginfo.fields.line = 0
     endif
 
+    " Make sure out 'end' is valid
```
alerque, line 1528:
```suggestion
    " Make sure our 'end' is valid
```

alerque (approved):
I looked through the code and it looks sane to me, but I haven't gotten a chance to test drive it. Use your judgement.

raven42:
So I've noticed some performance issues with this. When moving the cursor around in a larger file in an 'unscoped' area, the cursor movement is slower now because instead of finding a tag near to the cursor, it ends up looping through all tags and doesn't find one. I'm wondering if anybody knows of a better way to try to find a match instead of looping through all tags above the current cursor location looking for a matched scope? This issue could arise even when we are inside of a scope for something like a class that has a large number of functions defined in its declaration. Then when moving the cursor around in the 'class' scope toward the bottom, but outside of any member functions it would result in this same behavior.

raven42:
For now until any performance issues can be sorted out, I've changed the functionality so it has to be explicitly enabled by using the `'scoped-stl'` argument as a search-method for the `tagbar#currenttag()` and `tagbar#GetTagNearLine()` routines. These routines will default to the `'nearest-stl'` search method which is what they do today, but the user can optionally use the new search method to make it more intelligent at the cost of some performance.

alerque (approved):

alerque (approved):

PR #698, opened 2020-10-30
Add tag datatype

### LINKED ISSUE #680: show type of field by default
by pressing space key, the user can see the full prototype of field and methods, is it possible for tag bar to show the full prototype by default without the user pressing space? Thank you

raven42:
Closes #680

Add `g:tagbar_show_data_type` field to show the tag datatype next to the tag in the tagbar window
This uses the `--fields=t` field to get the datatype from ctags. If not found, then it will attempt to derive the datatype by extracting all the output from the `pattern` preceeding the tag name.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1342,7 +1342,7 @@ function! s:ExecuteCtagsOnFile(fname, realfname, typeinfo) abort
                           \ '-',
                           \ '--format=2',
                           \ '--excmd=pattern',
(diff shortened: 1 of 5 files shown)


raven42:
Ready to merge this in. As this is a new option that has to be explicitly enabled, I think we can do more testing with other languages and such after initial commit. In my testing with c / c++ files it seems to be stable.

alerque (approved):

PR #701, opened 2020-11-02
shane: tag line# at right may mess with type syn hl

Shane-XB-Qian:
refine for #687
there was perhaps a bit flaw, if set tag line# on and line# showed at right, looks it took same 'syn highlight' like 'parameter' or 'return type' etc such strings of ori func/tag; this is regardless compact or not, should be not an issue, but just maybe would not look very nice...
image
// try to fix it...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/doc/tags
+tags
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -1101,6 +1101,9 @@ TagbarScope
 TagbarType
     The type of a tag or scope if available.
 
+TagbarTagLineN
+    The source line number displayed to the right of each tag entry.
+
 TagbarSignature
     Function signatures.
 
--- a/syntax/tagbar.vim
+++ b/syntax/tagbar.vim
@@ -36,7 +36,8 @@ syntax match TagbarHelpKey   '" \zs.*\ze:' contained
 syntax match TagbarHelpTitle '" \zs-\+ \w\+ -\+' contained
 
 syntax match TagbarNestedKind '^\s\+\[[^]]\+\]$'
-syntax match TagbarType       ' : \zs.*'
+syntax match TagbarType       ' : \zs.*' contains=TagbarTagLineN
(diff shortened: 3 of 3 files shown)


raven42 (dismissed):

Shane-XB-Qian:
some langs e.g 'golang' had type like '[123]byte', so needed to refine a bit to get rid of such case from 'tag line# showed at right'.
// btw: but/though looks original code for 'syn hl' of 'type' etc itself seems did not cover all cases, e.g '[123]byte' was not highlighted.

// but: it's not regard to this pr, and not an issue, just may looks a bit not uniformed (in 'type' kind scope)..

raven42 (dismissed):

Shane-XB-Qian:
// adding the doc for the hl of TagbarTagLineN
looks that's what all needed to be added for this pr, @raven42 pls review and merge if ok. thx.

.gitignore
```diff
@@ -1 +1,2 @@
-/doc/tags
+# /doc/tags
```
alerque, line 1: Lets not comment out obsolete stuff, just delete it.

```suggestion
```

doc/tagbar.txt
```diff
@@ -1101,6 +1101,9 @@ TagbarScope
 TagbarType
     The type of a tag or scope if available.
 
+TagbarTagLineN
+    The line num of a tag if set it on at right.
```
alerque, line 1105: This needs some English copy editing.

```suggestion
    The source line number displayed to the right of each tag entry.
```

alerque (changes requested):

.gitignore
Shane-XB-Qian: that's comment as well, since not all people use 'tags', but just tags at current tag should had include 'doc' folder already
// but that's not a big deal, if you insisted, i can remove that/such 'comment'.

doc/tagbar.txt
Shane-XB-Qian: um... ok, i am not native english speaker, if you felt that's better, then ok.

alerque (approved):

alerque:
Thanks for the contribution.

PR #705, opened 2020-11-06
Add g:tagbar_jump_lazy_scroll option.

### LINKED ISSUE #703: New Option: g:tagbar_jump_lazy_scroll
Currently when jumping to a tag, the buffer will always be scrolled to place the tag at the location dictated by `g:tagbar_jump_offset`.  This means that if I jump to a tag that is already visible in the window (say, a few lines below my cursor) then the window will be scrolled unnecessarily, which can be a bit disorienting.

Solution: have a new option called `g:tagbar_jump_lazy_scroll` (defaults to `0`) which, when turned on, will cause the following jump behavior:

  1. If the tag being jumped to is already visible in the window then just put the cursor on that line without scrolling the window.
  2. If the tag is not visible in the window, then perform the current jumping/scrolling behavior whereby the buffer is scrolled to put the tag/cursor on the line dictated by `g:tagbar_jump_offset`.

I think this would make for a smoother and more intuitive user experience.

dpacbach:
When this option is on, a jump to a tag will only cause the
window to scroll if the tag line is not already visible on
the window.  If it is visible, the cursor will simply move
to that line without scrolling the window.  If the tagline
is not visible then the window will be scrolled as in current
behavior (according to the `g:tagbar_jump_offset` option).

Tested with and without the `g:tagbar_jump_offset` option set.

Fixes #703

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2286,6 +2286,12 @@ function! s:JumpToTag(stay_in_tagbar) abort
     " Mark current position so it can be jumped back to
     mark '
 
+    " Check if the tag is already visible in the window.  We must do this
+    " before jumping to the line.
+    let topline = line('w0')
+    let bottomline = line('w$')
+    let alreadyvisible = (taginfo.fields.line >= topline) && (taginfo.fields.line <= bottomline)
+
     " Jump to the line where the tag is defined. Don't use the search pattern
     " since it doesn't take the scope into account and thus can fail if tags
     " with the same name are defined in different scopes (e.g. classes)
@@ -2320,22 +2326,26 @@ function! s:JumpToTag(stay_in_tagbar) abort
         let taginfo.fileinfo.fline[curline] = taginfo
     endif
 
-    " Center the tag in the window and jump to the correct column if
-    " available, otherwise try to find it in the line
-    normal! z.
-
-    " If configured, adjust the jump_offset and center the window on that
-    " line. Then fall-through adjust the cursor() position below that
-    if g:tagbar_jump_offset != 0 && g:tagbar_jump_offset < curline
-        if g:tagbar_jump_offset > winheight(0) / 2
-            let jump_offset = winheight(0) / 2
-        elseif g:tagbar_jump_offset < -winheight(0) / 2
-            let jump_offset = -winheight(0) / 2
-        else
-            let jump_offset = g:tagbar_jump_offset
-        endif
-        execute curline+jump_offset
+    if g:tagbar_jump_lazy_scroll != 0 && alreadyvisible
+        " Do not scroll.
+    else
+        " Center the tag in the window and jump to the correct column if
+        " available, otherwise try to find it in the line
         normal! z.
+
+        " If configured, adjust the jump_offset and center the window on that
+        " line. Then fall-through adjust the cursor() position below that
+        if g:tagbar_jump_offset != 0 && g:tagbar_jump_offset < curline
+            if g:tagbar_jump_offset > winheight(0) / 2
+                let jump_offset = winheight(0) / 2
+            elseif g:tagbar_jump_offset < -winheight(0) / 2
+                let jump_offset = -winheight(0) / 2
+            else
+                let jump_offset = g:tagbar_jump_offset
+            endif
+            execute curline+jump_offset
+            normal! z.
+        endif
     endif
 
     if taginfo.fields.column > 0
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2269,6 +2269,15 @@ function! s:HighlightTag(openfolds, ...) abort
     endtry
 endfunction
 
+" Is the given line number already visible in the window without
+" any scrolling?
+function! s:IsLineVisible(line) abort
+    let topline = line('w0')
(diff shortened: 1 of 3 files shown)


autoload/tagbar.vim
```diff
@@ -2286,6 +2286,12 @@ function! s:JumpToTag(stay_in_tagbar) abort
     " Mark current position so it can be jumped back to
     mark '
 
+    " Check if the tag is already visible in the window.  We must do this
+    " before jumping to the line.
+    let topline = line('w0')
+    let bottomline = line('w$')
+    let alreadyvisible = (taginfo.fields.line >= topline) && (taginfo.fields.line <= bottomline)
```
alerque, line 2293: This logic looks fine, but I would like to request te abstract it to a new function that just returns a boolean answer for whether a line is visible. I suspect in the future we'll get a real API for this and it would be easier to swap out (and also use other places in our code) if it was abstracted as a function.

Additionally this would mean it could be called in the `if` statement *after* the other check, hence we can avoid the overhead of running this check if we're not going to use the answer anyway.

autoload/tagbar.vim
```diff
@@ -2320,22 +2326,26 @@ function! s:JumpToTag(stay_in_tagbar) abort
         let taginfo.fileinfo.fline[curline] = taginfo
     endif
 
-    " Center the tag in the window and jump to the correct column if
-    " available, otherwise try to find it in the line
-    normal! z.
+    if g:tagbar_jump_lazy_scroll != 0 && alreadyvisible
```
alerque, line 2329: Change this to call the fore mentioned function (hence only running if the option is set).

alerque (changes requested):
Thanks for taking the time to contribute. This looks pretty sound to me, I'd just suggest a small optimization.

autoload/tagbar.vim
dpacbach: Sure I can put it into a function.  However, I believe that we cannot call it in the `if` statement because, by the time we get there, the cursor has already been moved by the line:

```
execute taginfo.fields.line
```

This I believe will jump to the tag line, and in doing so, will change the values of the `topline` and `bottomline`, hence we need to call the function before this line.  Does that make sense?

autoload/tagbar.vim
dpacbach: See comment above for why we have to call the function above and cannot call it in the `if` statement.

autoload/tagbar.vim
dpacbach: If you like, for optimization purposes, I can just avoid calling the function unless the option is on, e.g.:

```vim
let noscroll = 0
if g:tagbar_jump_lazy_scroll != 0
  let noscroll = call s:is_line_already_visible(taginfo.fields.line)
endif
```

then later down in the `if` statement we only need to check `if noscroll != 0`.

alerque: On review I think you may be right about the logic here (although I didn't try to code it up the other way to proove the case). Let's go ahead with at least the small Optimization of checking whether the options is set before we do the extra math. With that done as you suggested here I think this will be ready for merge.

dpacbach: Done

dpacbach:
This PR is not showing my latest commit... do you know why?

dpacbach:
This PR is not showing my latest commit... do you know why?

alerque:
If you added more commits locally (to the branch `lazy-jump` that this PR is connected to) then you need to push them.

raven42:
did the push to github work? It might be asking for a merge in of the latest since there was another PR that just merged in.

dpacbach:
Yes the push seems to have worked, I can see the new commit in my repo: https://github.com/dpacbach/tagbar/commits/lazy-jump at the tip of the `lazy-jump` branch.

raven42:
Well that is interesting... I can see the second commit on the branch, and the parent has the correct hash. I've never seen where github doesn't automatically add that commit to the PR.

dpacbach:
Ok... well let me know if you think I should create a new PR and we close this one.

alerque:
I've never seen that either. Lets give it a few hours at least and see if it get unbroken on GitHub's side. It does look like it's been done correctly on your end, this seems like a caching or sharding bug on their end.

doc/tagbar.txt
```diff
@@ -1079,6 +1079,15 @@ Examples:
         " Set the tag jump location to appear 25% from the top
         let g:tagbar_jump_offset = winheight(0) / 4
 <
```
raven42, line 1081: There should be a jump target line that is right justified here just before the `~` line.
```
                                                                        *g:tagbar_jump_lazy_scroll*
g:tagbar_jump_lazy_scroll~
...
```

raven42 (changes requested):
Need a small addition to the help file.

doc/tagbar.txt
dpacbach: Done

dpacbach:
Looks like my new commits have now appeared!

raven42 (approved):

alerque (approved):

dpacbach:
Thanks; feel free to merge when ready.

PR #707, opened 2020-11-09
Add linenr protection into getPrototype

### LINKED ISSUE #706: Seeing `s:getPrototype: line 33: E684:list index out of range: 0' error when cursor positioned in tagbar window
On some filetypes (python in this instance), I am seeing the following error when moving the cursor around in the tagbar window. The loop in `s:getPrototype()` does not have protection for going beyond the file line limit.

```
Error detected while processing function <SNR>73_ShowPrototype[8]..<SNR>78_getPrototype:
line   33:
E684: list index out of range: 0
Press ENTER or type command to continue
```

The tag in question has an `=` character in it. When selecting a tag that does not have an `=` character works fine.

raven42:
Closes #706

Add protection to not go beyond the last line in the file when calling the `s:getPrototype()` routine. This occurs when there is an `=` character in the tag prototype. For example in a python tag with a parameter line this: `some_function(arg1, optional_arg_2=False)`

### DIFF
--- a/autoload/tagbar/prototypes/normaltag.vim
+++ b/autoload/tagbar/prototypes/normaltag.vim
@@ -104,7 +104,7 @@ function! s:getPrototype(short) abort dict
 
         let prototype = line
         let curlinenr = self.fields.line + 1
-        while balance > 0
+        while balance > 0 && curlinenr < line('$')
             let curline = getbufline(bufnr, curlinenr)[0]
             let curlist = split(curline, '\zs')
             let balance += count(curlist, '(')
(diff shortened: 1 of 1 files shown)


raven42:
This issue is a side effect of #698 for #680.

Fixing this issue may cause the 'auto' datatype calculation for kinds which ctags does not report the tag datatype to also include any declaration information.

The intent of the `let line = substitute(line, '\s*=.*', '', '')` line in normaltag.vim was to handle cases where a tag was defined like this (example in Java):
```java
public class some_class {
    public status final String SOME_STRING = "some string value";
}
```

At this time, universal ctags does not interpret the `--fields=-t` option for java files and does not provide the datatype. So there is an attempt by tagbar to figure out what the datatype is. This line was included to strip off the declaration from the prototype as well in an effort to clean it up. However this now causes a bug for python files which can include a declaration of optional variables inside of the prototype. Hence we have to remove this line until something else can be figured out.

This will have the downside of any files which ctags does not know how to provide the `datatype` field for, they will get the full datatype including any declarations of the tag variable.

alerque (approved):

PR #709, opened 2020-11-11
Add option for g:tagbar_highlight_method

### LINKED ISSUE #708: regression issue started from pr #683
@raven42  there is a regression issue started from pr #683 : tag highlight does not work Unless cursor on that func name same line.
image
// so far not sure which piece caused this, but looks started from this pr.
// could you check to refine the logic, or revert, or had a option to disable it ? thx.

raven42:
Closes #708

Add option for g:tagbar_highlight_current_tag which will allow
highlighting of tags on the current line in addition to highlighting the
scoped tags. Reverts the behavior added in #683 to the default from
before that commit

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -518,9 +518,11 @@ function! tagbar#types#uctags#init(supported_types) abort
     \ ]
     let type_go.sro = '.'
     let type_go.kind2scope = {
+        \ 'p' : 'packages',
         \ 's' : 'struct'
     \ }
     let type_go.scope2kind = {
+        \ 'packages' : 'p',
         \ 'struct' : 's'
     \ }
     let types.go = type_go
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -518,11 +518,9 @@ function! tagbar#types#uctags#init(supported_types) abort
     \ ]
     let type_go.sro = '.'
     let type_go.kind2scope = {
-        \ 'p' : 'packages',
         \ 's' : 'struct'
     \ }
     let type_go.scope2kind = {
-        \ 'packages' : 'p',
         \ 'struct' : 's'
     \ }
     let types.go = type_go
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -1090,6 +1090,33 @@ containing the tag without scrolling the window.  If the tag is not visible
 in the window then the window will be scrolled and the tag (and cursor)
 placed in the location dictated by |g:tagbar_jump_offset|.
 
+                                              *g:tagbar_highlight_current_tag*
+g:tagbar_highlight_current_tag~
+Default: 0
+
+If set to non-zero, then tag highlighting will also highlight any tags located
+on the same line as the cursor. By default only the nearest tag with the {stl}
+attribute enabled will be highlighted.
+
+For example in the following C example, if the cursor is on the
+'#define SOME_MACRO' line, without this option set, which is the default
+behavior, then the tag 'some_function()' will be highlighted.  With this option
+set to non-zero, then that macro tag for 'SOME_MACRO' will be highlighted while
+the cursor is on that line. But then will revert to highlighting the
+'some_function()' tag when the cursor is not on that line.
+>
+        int some_function(int arg) {
+            int var1;
+            int var2;
+        #define SOME_MACRO 1
+            ...
+            printf("...");
+        }
+<
+Example: >
+        let g:tagbar_highlight_current_tag = 1
+<
+
 ------------------------------------------------------------------------------
 HIGHLIGHT COLOURS                                           *tagbar-highlight*
 
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -1134,7 +1134,7 @@ highlight the 'SOME_MACRO' tag.
         15 }
 <
 Example: >
-        let g:tagbar_highlight_current_tag = 'nearest'
+        let g:tagbar_highlight_method = 'nearest'
 <
 
 ------------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


raven42:
Reworked to use a different configuration of `g:tagbar_highlight_method` instead which can be set to one of `'nearest'` / `'nearest-stl'` / `'scoped-stl'` to highlight the appropriate tag.

autoload/tagbar/types/uctags.vim
```diff
@@ -518,9 +518,11 @@ function! tagbar#types#uctags#init(supported_types) abort
     \ ]
     let type_go.sro = '.'
     let type_go.kind2scope = {
+        \ 'p' : 'packages',
```
Shane-XB-Qian, line 521: looks not work..

doc/tagbar.txt
```diff
@@ -1090,6 +1090,53 @@ containing the tag without scrolling the window.  If the tag is not visible
 in the window then the window will be scrolled and the tag (and cursor)
 placed in the location dictated by |g:tagbar_jump_offset|.
 
+                                                  *g:tagbar_highlight_method*
+g:tagbar_highlight_method~
+Default: 'nearest-stl'
```
Shane-XB-Qian, line 1095: i wish the default is 'nearest' which kept previous default if so.

doc/tagbar.txt
```diff
@@ -1090,6 +1090,53 @@ containing the tag without scrolling the window.  If the tag is not visible
...
+        9      }
+        10
+        11     int another_function(int arg) {
+        12         int varA;
+        13         printf("###");
+        14     }
+        15 }
+<
+Example: >
+        let g:tagbar_highlight_current_tag = 'nearest'
```
Shane-XB-Qian, line 1137: option name is 'g:tagbar_highlight_method'

plugin/tagbar.vim
```diff
@@ -100,6 +100,7 @@ function! s:setup_options() abort
         \ ['jump_lazy_scroll', 0],
         \ ['left', 0],
         \ ['help_visibility', 0],
+        \ ['highlight_method', 'nearest-stl'],
```
Shane-XB-Qian, line 103: i wish the default is 'nearest' which kept previous default if so.

autoload/tagbar/types/uctags.vim
```diff
@@ -518,9 +518,11 @@ function! tagbar#types#uctags#init(supported_types) abort
...
         \ 's' : 'struct'
     \ }
     let type_go.scope2kind = {
+        \ 'packages' : 'p',
```
Shane-XB-Qian, line 525: looks not work..

autoload/tagbar/types/uctags.vim
raven42: I'll leave this one out.

doc/tagbar.txt
raven42: I'm really hesitant to default to `'nearest'` as that would be the incorrect behavior for most users. In all my testing, the `.stl` flag has been set correctly even in `.go` files, so I would want to see the default behavior be what most users will expect the behavior to be. Having the option to override to `'nearest'` allows the flexibility so the user can change it if they want. It appears even in your own testing that the `.stl` flag is correct in some cases, and incorrect in others. That is likely due to some type of environmental setup difference between how the tagbar configurations are loaded or ctags output is generated. Without a way to reproduce that, I can't dig into it further, so unless many other people are seeing this behavior too, then defaulting to `'nearest-stl'` should be the expected behavior.

doc/tagbar.txt
raven42: good catch

plugin/tagbar.vim
raven42: see comment from above

autoload/tagbar/types/uctags.vim
raven42: will remove this

alerque (approved):

doc/tagbar.txt
Shane-XB-Qian: i think put 'nearest' to be the default is following the general best practice which would not against previous default, and here actually not regard to 'correct' or 'incorrect', since which/where 'tag' to be highlighted actually was a personal preference..
but anyway, now we had a safeguard option, that's a bit better already. So up to you. thx.

Shane-XB-Qian:
actually now, even 'nearest' is better than previous 'default', if current line no tag then it will highlight nearest above one.
https://github.com/preservim/tagbar/issues/708#issuecomment-725459627
"to me, to show the parent func name (if cursor in that scope) is more important than show tag in same line."
// we did it now. thx buddy @raven42 we've work this out. :+1:

raven42:
Glad we could get things working for you again.

Shane-XB-Qian:
for everybody actually :+1:
// #683 was hardcode to 'nearest-stl' only.. now, even 'nearest' looks is better.
for example, even for 'c' (vs 'golang'), following c code, if cursor was Not on a macro (on its next line), it was hightlighted nothing, now with 'nearest', it can hightlight nearest above one. should be more make sense now.. :)
```c
static char *(main_errors[]) =
{
    N_("Unknown option argument"),
#define ME_UNKNOWN_OPTION	0
    N_("Too many edit arguments"),
#define ME_TOO_MANY_ARGS	1
    N_("Argument missing after"),
#define ME_ARG_MISSING		2
    N_("Garbage after option argument"),
#define ME_GARBAGE		3
    N_("Too many \"+command\", \"-c command\" or \"--cmd command\" arguments"),
#define ME_EXTRA_CMD		4
    N_("Invalid argument for"),
#define ME_INVALID_ARG		5
};
```

Shane-XB-Qian:
and `.stl` still was an issue actually, and not only for 'golang'..
// different ctags_bin configured different value in different places, and ctags_bin auto detection override the configure one.. 
// looks such situation would make 'nearest-stl' (which depeneds on `.stl`) behaviour a little flaky..
anyway/so yes, i preferred 'nearest', wish it kept working always..

PR #712, opened 2020-11-20
Consistent Documentation: Tagbar.txt now has same example TagbarToggle hotkey as README.md

pascalbakker:
README TagbarToggle example uses F8 but the tagbar.txt example uses F9.


alerque (approved):
Thanks for taking a moment to contribute towards better docs.

PR #722, opened 2020-12-21
Add mapping for arduino filetypes

### LINKED ISSUE #721: Remap Filetypes
Is it possible to make it accept a filetype as if it was another? Like I have an extension that provides highlight and autocomplete for arduino files which I can even generate tags for but tagbar will be empty when i open if If i set ft to c tagbar works.

Can I remap those filetypes somehow?

raven42:
Closes #721

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -250,6 +250,7 @@ function! tagbar#types#uctags#init(supported_types) abort
     \ }
     let types.cpp = type_cpp
     let types.cuda = type_cpp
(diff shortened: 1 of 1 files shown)


alerque (approved):
LGTM.

Is there a canonical place we document filetype support?

raven42:
> LGTM.
> 
> Is there a canonical place we document filetype support?

Doesn't look like it. I did a search for the `cuda` filetype which is also linked to `c++` in a similar way and did not find anything other than the actual definition in uctags.vim and ctags.vim.

PR #725, opened 2020-12-30
Add `g:tagbar_highlight_follow_insert` option

### LINKED ISSUE #724: Tagbar highlight not following the cursor in insert mode
Moving the cursor in the main buffer causes the relevant symbol/section to be automatically highlighted in the tagbar in normal mode, but not in insert mode, even though it would be useful in this case too.

A workaround for this is to set the following autocmd:
```
autocmd CursorHoldI * TagbarShowTag
```

However, this workaround generates warning messages when the tagbar is closed.

pacien:
GitHub: closes #724

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -576,6 +576,10 @@ function! s:CreateAutocommands() abort
         " was changed by an external command; see commit 17d199f
         autocmd BufReadPost,BufEnter,CursorHold,FileType * call
                     \ s:AutoUpdate(fnamemodify(expand('<afile>'), ':p'), 0)
+        if g:tagbar_highlight_follow_insert
+            autocmd CursorHoldI * call
+                        \ s:AutoUpdate(fnamemodify(expand('<afile>'), ':p'), 0)
+        endif
         autocmd BufDelete,BufWipeout *
                     \ nested call s:HandleBufDelete(expand('<afile>'), expand('<abuf>'))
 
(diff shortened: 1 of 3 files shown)


raven42 (approved):
Looks good to me. Do the extra file system calls to check if the file exists and file modify time pose any problem if the `updatetime` is shorter and there are pauses between editing?

raven42:
Hmm... if there is indeed a lag introduced that might not be good. In other modes we just have to worry about the user moving the cursor around. But in insert mode, the user could also be writing code or something and if there is a brief pause there, then that could interfere with typing.

I would suggest we add an option to enable this instead then instead of just globally enabling it. Then you can enable it in your environment without impacting others.

raven42 (changes requested):
Since there is a delay introduced, this should be enabled under a new tagbar option such as `g:tagbar_insert_mode_udpate` or similar.

alerque (changes requested):
Per both discussion in the PR and my comments on #724, this does come with a very big potential for performance issues. Insert mode is particularly problematic, and no processing should be enabled by default in insert mode that isn't strictly required for function. The UX when insert mode hangs for some plugin action that isn't essential is very frustrating. This is especially a problem on larger files or with pastes, etc.

I'd be happy to see this supported, but it *must* be behind a feature flag that defaults to off so people that want it can pick their poison without degrading the experience for everyone out of the box.

pacien:
It's now disabled by default and guarded by the `g:tagbar_highlight_follow_insert` option.

alerque (approved):
Thanks for making those changes. I think this looks good now...

raven42 (approved):

PR #733, opened 2021-01-05
Use universal_newlines for Popen call

### LINKED ISSUE #629: [windows] when tagbar_python is enabled, tag identifiers contains ^M
Can be fixed by adding `universal_newlines=True` to the arguments of `Popen`.

```diff
diff --git a/autoload/tagbar.vim b/autoload/tagbar.vim
index 0c2c42e..d14f909 100644
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2892,7 +2892,7 @@ function! s:run_system(cmd, version) abort
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
-        exec pyx . '__pp = subprocess.Popen(**__argv)'
+        exec pyx . '__pp = subprocess.Popen(**__argv, universal_newlines=True)'
         exec pyx . '__return_text = __pp.stdout.read()'
         exec pyx . '__pp.stdout.close()'
         exec pyx . '__return_code = __pp.wait()'
```

raven42:
Closes #629

Not actually tested on windows as I don't have a windows installation to test with. Will be a good idea if somebody can verify on windows.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3030,7 +3030,7 @@ function! s:run_system(cmd, version) abort
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
-        exec pyx . '__pp = subprocess.Popen(**__argv)'
+        exec pyx . '__pp = subprocess.Popen(**__argv, universal_newlines=True)'
         exec pyx . '__return_text = __pp.stdout.read()'
         exec pyx . '__pp.stdout.close()'
         exec pyx . '__return_code = __pp.wait()'


alerque (approved):

alerque:
This looks sane to me, but I don't actually have Windows to test with either. Confirmation from *anyone* would be appreciated, although if this sits too long without feedback I'd be okay with merging anyway and see if anybody complains.

CC @char101

char101:
I have been using that modification since the issue is created on Windows, regularly rebased and haven't found any problem. If any there could be problem with python2, if vim is only linked with python2.

raven42:
According to the python 2.7 documentation, the Popen constructor does still accept `universal_newlines=True` so it looks like we'll be ok there as well.

Thanks for the verification @char101

skywind3000:
You forgot to add a `errors = "ignore"` and popen may crash when filenames contain strange CJK characters, see:

https://github.com/preservim/tagbar/pull/745

PR #734, opened 2021-01-06
Fix the problem that the data type was unexpectedly displayed in the C++ destructor's tag.

dan0907:
Make some corrections to  #732 
Same as constructor, destructor has no data type
Solve the problem shown in the figure below

擷取

### DIFF
--- a/autoload/tagbar/prototypes/normaltag.vim
+++ b/autoload/tagbar/prototypes/normaltag.vim
@@ -152,7 +152,7 @@ function! s:getDataType() abort dict
         endif
 
         let line = getbufline(bufnr, self.fields.line)[0]
-        if (self.name =~ '^\s*\~')
+        if (self.name =~# '^\s*\~')
             let data_type = ''
         else
             let data_type = substitute(line, '\s*' . self.name . '.*', '', '')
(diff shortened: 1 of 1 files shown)


raven42:
Looks good to me. I left the `~` in intentionally as I thought it added something to know that as the datatype, but then again, I don't use c++ much, so I may not be the best one to make that call. If this is the preferred appearance then it looks good.

autoload/tagbar/prototypes/normaltag.vim
```diff
@@ -152,8 +152,11 @@ function! s:getDataType() abort dict
         endif
 
         let line = getbufline(bufnr, self.fields.line)[0]
-        let tmp_name = substitute(self.name, "\\~", '', 'g')
-        let data_type = substitute(line, '\s*' . tmp_name . '.*', '', '')
+        if (self.name =~ '^\s*\~')
```
raven42, line 155: As per VINT checks, this should be `=~#` instead of `=~`. Otherwise looks good.

raven42 (approved):

autoload/tagbar/prototypes/normaltag.vim
alerque, line 155:
```suggestion
        if (self.name =~# '^\s*\~')
```

alerque (approved):
Thanks for taking the time to contribute @dan0907.

I fixed the lint error but am not in a position to actually test drive this. @raven42 if it looks good to you I'd say go ahead and merge (via squash).

PR #735, opened 2021-01-10
Improve tag highlighting and signature matching

dan0907:
1. Correctly highlight tags that contain spaces.
2. Match function signature on special case: operator()(...).

e.g.
sig
tag

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2273,10 +2273,11 @@ function! s:HighlightTag(openfolds, ...) abort
 
         " If printing the line number of the tag to the left, and the tag is
         " visible (I.E. parent isn't folded)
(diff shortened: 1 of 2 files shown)


raven42 (approved):

Shane-XB-Qian:
@dan0907 @raven42 
`E488: Trailing characters: //en.wikipedia.org/wiki/foo)\m\ze/: match TagbarHighlight /^\%28l\s*[▼▶ ][-+# ]\?\zs\VSource: [Wikipedia](https://en.wikipedia.org/wiki/foo)\m\ze/`

raven42:
@dan0907 Would you be able to look into this? My knowledge of highlight search patterns is quite limited. If need be we can revert these changes to restore original functionality until this can be addressed.

Issue #736, opened 2021-01-12
E488: Trailing characters for match TagbarHighlight - Seen when viewing a tag with a hyperlink


raven42:
Introduced in PR #735. See https://github.com/preservim/tagbar/pull/735#issuecomment-758689658 for more info.

PR #737, opened 2021-01-12
Escape special characters in pattern searching

dan0907:
@raven42 @Shane-XB-Qian 
This should resolve the bug #736

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2288,7 +2288,7 @@ function! s:HighlightTag(openfolds, ...) abort
 
         " If printing the line number of the tag to the left, and the tag is
         " visible (I.E. parent isn't folded)
-        let identifier = '\zs\V' . tag.name . '\m\ze'
+        let identifier = '\zs\V' . escape(tag.name, '/\') . '\m\ze'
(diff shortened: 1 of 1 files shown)


alerque (approved):

raven42 (approved):

Shane-XB-Qian:
Actually I had no confidence on your change on #735 the ‘pattern’, and the ‘hl’ a bit too specific to a special case, looks...
As for this md link header tag issue, just a case, I will try later..

alerque:
Thanks for the quick fix @dan0907.

@Shane-XB-Qian If there is something still the matter feel free to report it, or even better submit a superior alternative, PRs are welcome for anything you know how to improve upon.

Shane-XB-Qian:
@alerque donot need to say this.. you're admin-ing this repo, should be awared tagbar is not only working for  one/two specific langs.. better to avoid 'magic chars(/numbers)'..
i supposed @dan0907 and/or @raven42 knew what i mean..

Shane-XB-Qian:
@dan0907 @raven42 to this issue itself (md link header tag), i tried, looks fixed. thx.

PR #738, opened 2021-01-14
add a new language

lasorda:


raven42 (changes requested):
This should probably also be added to uctags.vim as well. ctags.com is for exuberant ctags which is no longer supported. So anything being added to ctags.vim should also be included in uctags.vim.

alerque (approved):
Thanks for the contribution. As far as I'm aware this should be good enough. Since we don't have any canonical list of supported filetypes to update anywhere...

raven42 (approved):

PR #745, opened 2021-01-23
fixed: CJK characters decoding error problem

skywind3000:
[PR 733](https://github.com/preservim/tagbar/pull/733) changed popen output format from binary to text but forgot to initialize `errors` to `ignore`, which may crash when there are unicode decoding errors:

图片

and:

图片

When filenames contain strange CJK characters, python's `decode` may fail. A `errors = "ignore"` can simply fix this.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3051,7 +3051,7 @@ function! s:run_system(cmd, version) abort
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
-        exec pyx . '__argv["errors"] = "ignore" '
+        exec pyx . '__argv["errors"] = "ignore"'
         exec pyx . '__pp = subprocess.Popen(**__argv, universal_newlines=True)'
         exec pyx . '__return_text = __pp.stdout.read()'
         exec pyx . '__pp.stdout.close()'
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3051,6 +3051,7 @@ function! s:run_system(cmd, version) abort
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
+        exec pyx . '__argv["errors"] = "ignore" '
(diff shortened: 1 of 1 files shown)


skywind3000:
@alerque , how to request a review ?

alerque (approved):

autoload/tagbar.vim
```diff
@@ -3051,6 +3051,7 @@ function! s:run_system(cmd, version) abort
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
+        exec pyx . '__argv["errors"] = "ignore" '
```
alerque, line 3054:
```suggestion
        exec pyx . '__argv["errors"] = "ignore"'
```

Is that a stray space at the end of this string? I assume this array gets joined or passed as an array and doesn't need trailing spaces, no?

alerque:
Thanks for the contribution. This is slightly outside my expertise in Python, but the fix looks reasonable and I'll take your word for it being the right fix. Just one possible nit pick on the coding...

alerque (approved):

Shane-XB-Qian:
pls check `let g:tagbar_systemenc = 'gbk'` or sth,
or perhaps better try to fix the why/failure of encoding vs (not to) 'ignore' ...................... i think!

skywind3000:
`let g:tagbar_systemenc = 'gbk'` won't help, the binary output is decoded from binary to text in python after PR733.

python's `subprocess.Popen` module can detect default system encoding and use `gbk` for decoding.

See the error message above:

    gbk codec can't decode byte 0xae in position 680

Popen has already detected the system encoding as 'gbk' correctly, the issue is not caused by wrong codec name.

So, `let g:tagbar_systemenc` os similar things will not help.

skywind3000:
Python's gbk decoder uses very strict rules and there may be slight differece with windows gbk/gb2312 codec. 

Python raise an exception when unicode decoding errors encountered. **It is not an error**, but a chance to allow users to decide what to do with the exception characters (simply ignore or replace it with a `?`).

Most of time, set `errors="ignore"` is a practical way.

Shane-XB-Qian:
it's not regarding to codec name, but 'why', perhaps his/its ori src code was wrong (codec) too..
so i think perhaps better to fix the why..
// and try `let g:tagbar_python = 3` or `let g:tagbar_python = 2`

skywind3000:
No sense to argue with the codec, filename field in the ctags database is neally unused in tagbar.

So, this PR is enough for this. You can still create your new patch if you don't agree with this.

PR #752, opened 2021-02-12
Fix incorrect example for g:tagbar_status_func

maujim:
Arguments being accessed in the function body weren't prefixed with `a:` which caused the example to fail.


raven42 (approved):

raven42:
Thanks for the contribution!

PR #761, opened 2021-05-14
README: Name the recommended ctags impl first

shuber2:


README.md
```diff
@@ -23,11 +23,12 @@ creates the tags it needs on-the-fly in-memory without creating any files.
...
-  or (**highly recommended**) any version of [Universal
-  Ctags](https://ctags.io) which is a currently maintained fork of Exuberant
-  Ctags with many bugfixes, support for many more formats, and proper Unicode
-  support. Some additional formats can also be handled by other providers such
+* A ctags implementation: We _highly recommend_ any version of [Universal
+  Ctags](https://ctags.io). It is a maintained fork of Exuberant Ctags with
+  many bugfixes, support for many more formats, and proper Unicode support.
+
+  Alternatively, use [Exuberant Ctags](http://ctags.sourceforge.net/) 5.5 or
+  higher. Some additional formats can also be handled by other providers such
```
alerque, line 31:
```suggestion
  Alternatively, use the [Exuberant Ctags](http://ctags.sourceforge.net/) 5.5 or
  higher. Note this supports less formats and will eventually be deprecated.
  
  Some additional formats can also be handled by other providers such
```

alerque (changes requested):
I like it, but I suggest being even more aggressive steering people away from Exuberant. We've pretty much said we don't care about supporting it any more. We haven't broken it on purpose but we're not doing much work for it and do plan to eventually drop it entirely.

README.md
shuber2: Yes, makes sense. What about this version?

alerque (approved):
Thanks for taking the time to contribute!

PR #764, opened 2021-05-19
Screenshot add with Go language

vitaly-zdanevich:


raven42:
My personal opinion is we should not add screenshots to the repository itself. Any screenshots should be hosted either by something like imgur like the existing screenshots, or some other method of hosting. It is possible to upload the screenshot as part of an issue under a repository, and then use the raw image link to provide the link. I can't quite remember exactly how to do that though.

Also I believe we should try to format the screenshot the same as the others for uniformity. At a minimum it should be the same size, aspect ratio, and font size as the other screenshots so it looks and feels the same.

If we do want to add this to the repository itself, then there should be an `/assets/` directory created for things like that.

Just my 2 cents.

alerque:
I completely agree with @raven42 here, I do not want to see screenshots going into the repository itself. They are dead weight that does not need to be cloned into thousands of people's plugin directories and they are very prone to becoming out of date.

I'm not so opposed to seeing images embedded in the readme, but they need to be hosted here on Github (just insert an image into an issue, then use the Markdown coded it gives for the upload) or somewhere else central. I don't like imgurl for this because I don't really like the privacy leak it introduces of a third party seeing analytic data. GitHub is only less onerous in this respect because most people will see it rendered here anyway.

Lastly lets make user any screenshots we add do a good job of showcasing the plugin functions. We don't need one for every possible language, only one or two that showcase simple and advanced looking configurations.

alerque (changes requested):
Remove raster image from repo itself, possibly host as asset in GitHub issues. Better yet why not just put this in the Wiki? We could start a showcase there and let people add at will. At most I'd like to see one or two in the readme.

vitaly-zdanevich:
>  I do not want to see screenshots going into the repository itself. They are dead weight that does not need to be cloned into thousands of people's plugin directories

Sometimes I do not remember what this plugin do - this is convenient that I can see a screenshot right from the directory.

alerque:
> > I do not want to see screenshots going into the repository itself. They are dead weight that does not need to be cloned into thousands of people's plugin directories
> 
> Sometimes I do not remember what this plugin do - this is convenient that I can see a screenshot right from the cloned directory.

Most people use a plugin manager to maintain the clones of various plugins. These configs usually lend themselves to adding comments next to the load lines if you have trouble remembering what each plugin is for.

```viml
" Display and navigate tags in a sidebar
Plug 'preservim/tagbar'
```

Second, the README should serve this purpose much better, especially since it works in the terminal to review what a repository is about.

Third, it shouldn't take a third screen shot to serve this purpose. One is more than enough. In fact I think 1 in the readme is find, but even one in the repo may be too many. I have over 100 plugins and routinely sync my vim configs across an assortment of machines including low-resource embedded devices. The repositories that have images and other cruft are obnoxious.

PR #765, opened 2021-05-19
README: add about how to install tags on Gentoo

vitaly-zdanevich:


raven42:
I'm not sure if we want to call out specific linux distribution installation methods here. That kind of thing can change when new images come out. If we do want this, I would recommend a subsection heading for linux distribution package paths... something like this, though again, I see this being a maintenance issue trying to keep all the links up to date. And it doesn't really make sense to only provide one distro.

```markdown
## Dependencies

* [Vim](http://www.vim.org/) >= 7.3.1058
  or any version of [NeoVim](https://neovim.io/).

* A ctags implementation: We _highly recommend_ any version of [Universal
  Ctags](https://ctags.io). It is a maintained fork of Exuberant Ctags with
  many bugfixes, support for many more formats, and proper Unicode support.

  [Exuberant Ctags](http://ctags.sourceforge.net/) 5.5 or higher works to some
  degree but will be deprecated eventually.

  Some additional formats can also be handled by other providers such as
  [jsctags](https://github.com/sergioramos/jsctags) or
  [phpctags](https://github.com/vim-php/phpctags).

#### Linux Package Locations
 - Gentoo: [dev-util/ctags](https://packages.gentoo.org/packages/dev-util/ctags) (Universal CTAGS)
 - Other Distro: (other-links)

## Installation

Extract the archive or clone the repository into a directory in your
....
```

raven42:
I do appreciate the idea to help make it easier for users to install though. Just needs to be a little more comprehensive I think.

I'll let others decide though.

alerque (changes requested):
The distro package naming and versioning schemes for ctags are a nightmare. I'm afraid if we start calling them out here it's going to be a never-ending game to stay up to date and complete. Most people are going to know how to find packages for their distro of choice anyway, and for those that aren't we would have to be very verbose here to effectively help.

I suggest we just link to somewhere people can find their distro's package name, like for example Repology: https://repology.org/project/universal-ctags/versions

PR #769, opened 2021-05-26
Fix for #750 to cleanup files once tagbar window closed

### LINKED ISSUE #750: TagbarToggle and reopen file
Open File
Open Tagbar
Close Tagbar
Wipe File
Open File
Open Tagbar
Select a tag
Message:
Error detected while processing function <SNR>83_JumpToTag[11]..<SNR>83_GotoFileWindow:
line   11:
E86: Buffer 4 does not exist

Suggestion:
In autoload/tagbar.vim at the end of s:CloseWindow():
   
Add-> let s:known_files._files = {}
    
       call tagbar#debug#log('CloseWindow finished')
endfunction

raven42:
Fixes #750

Once the tagbar window is opened and a file is registered, that file
stays in memory even if the tagbar window is closed. This allows tagbar
to cache the info so if the tagbar window is opened again, it doesn't
have to rescan the file and rerun ctags on the file.

However if this buffer is wiped out using `:bwipe <file>`, then the
buffer is completely unloaded from vim memory and also needs to be
unloaded from tagbar memory. This works if the tagbar window is open,
but in the event the tagbar window is closed, all autocmds are
unregistered, so tagbar never gets the BufWipeout notification. This
results in tagbar leaving the buffer in active memory even though the
buffer doesn't exist anymore.

This fix will leave the BufWipeout and BufDelete autocmds active even if
the tagbar window is closed. This allows the buffer cleanup to occur on
a `:bwipe` command even if the tagbar window is closed.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -582,8 +582,6 @@ function! s:CreateAutocommands() abort
                 autocmd CursorHoldI * call
                         \ s:AutoUpdate(fnamemodify(expand('<afile>'), ':p'), 0)
             endif
-            autocmd BufDelete,BufWipeout *
-                        \ nested call s:HandleBufDelete(expand('<afile>'), expand('<abuf>'))
 
             " Suspend Tagbar while grep commands are running, since we don't want
             " to process files that only get loaded temporarily to search them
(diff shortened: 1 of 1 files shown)


hnordholz (approved):
Works as expected.

raven42:
Perfect. Thanks!

Issue #772, opened 2021-06-18
while split tagbar jumptotag fail 


wellcomez:
Error message w:autoclose undefined
function! s:JumpToTag(stay_in_tagbar) abort
    let taginfo = s:GetTagInfo(line('.'), 1)
    let autoclose= w:autoclose  
  

> let autoclose= exists('w:autoclose') && w:autoclose ? 1 : 0 

> check whether w:autoclose existed

raven42:
I have pushed up a potential fix. There are a lot more than just the `w:autoclose` that are not set. To properly fix this I think we'd need to call the `s:InitWindow()` when a split occurs. The PR can be used as a baseline, but I haven't fully tested everything or fully examined the implications. On basic testing it does appear to fix the issue though and the split works successfully after this patch.

PR #773, opened 2021-06-18
Potential fix for #772 - InitWindow issue

raven42:
Potential fix for #772

This issue is being caused by the InitWindow() routine not being called
when a split of the tagbar window occurs. This will potentially fix that
issue by forceably calling s:InitWindow() when a new tagbar window is
found.

Note however, that when closing one of the split tagbar windows, it does
cause the other one to close as well. It looks like there is an issue on
the close processing that it identifies all windows that have the tagbar
buffer loaded as the tagbar window.


raven42:
This is not a fully functional fix. After some testing I've found odd interactions with other plugins. Still need to look into why.

PR #775, opened 2021-06-22
Fix perl definitions

### LINKED ISSUE #774: [question, feature request] how to check that the tagbar plugin is loaded?
Seeminly there exists no such global boolean variable like `g:loaded_tagbar` or `g:loaded_tagbarPlugin` or `g:loaded_tagbar_plugin` to check that the tagbar plugin is loaded. What alternative exists? If no proper alternative exists, could one of the previous variables be implemented and documented?

raven42:
Closes #774

Pending verification.


alerque (approved):
I also don't work with Perl (any more, at least when I can help it), but especially if anybody that does test it with Perl signs off this looks fine to me.

raven42:
Fix working as per @XSven

PR #780, opened 2021-07-26
Add JumpToNearbyTag functionality

### LINKED ISSUE #774: [question, feature request] how to check that the tagbar plugin is loaded?

### LINKED ISSUE #779: Need help to understand how to implement jump to previous and next tag outside from tagbar window.
Hi,
Currently I want to have the following use case:
1) I want to open the tagbar without going into it's window.
2) I want to jump to next tag (variables, functions, etc) without going into tagbar's window. Expectation is to smoothly traverse the code and not the tagbar window.

The above functionality is similar to [[ and ]], however it works correctly only with C and C++ files. Java files tend to jump to top and bottom of the file and not to their respective functions and variables.

Is there a way to use tagbar's existing functionality so I can remap [[ ]] and move to next tag without going into tagbar window (and hence not closing it).

Thanks!

raven42:
Closes #779

Updated functions:

`s:GetNearbyTag()` - Allow for a direction parameter to be passed into the routine. This will allow for a forward or backward search. It will default to a backward search line the current behavior if the `a:direction` parameter is not used.

`s:JumpToTag()` - Allow for an optional parameter to be passed in to specify the tag instead of using the tag under the cursor on the current line. This also allows this routine to be executed from the file window instead of the tagbar window. Assume `autoclose=0` if passing in the tag.

Add new functions:

`s:JumpToNearbyTag()` - This routine will do a forward or backward search for the nearst tag using the `GetNearbyTag()` with the new direction field, then call the `JumpToTag()` with the tag we found.

`tagbar#jumpToNextTag()` - New public facing wrapper routine around the `JumpToNearbyTag()` routine. Optionally take in `a:lnum` and `a:search_method` to be used in other routines. If `a:lnum` is not present, then the line number will search from the next or previous line depending on the `a:direction` value.

TODO:
- [x] Still need to write up the documentation for this.
- [x] Possibly look at providing default keymap. Currently this can be done using a custom definition in `.vimrc`.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2328,17 +2328,8 @@ endfunction
 
 " s:JumpToTag() {{{2
 function! s:JumpToTag(stay_in_tagbar, ...) abort
-    if a:0 > 0
-        let taginfo = a:1
-    else
-        let taginfo = s:GetTagInfo(line('.'), 1)
-    endif
-
-    if a:0 > 1
-        let force_lazy_scroll = a:2
-    else
-        let force_lazy_scroll = 0
-    endif
+    let taginfo = a:0 > 0 ? a:1 : s:GetTagInfo(line('.'), 1)
+    let force_lazy_scroll = a:0 > 1 ? a:2 : 0
 
     if empty(taginfo) || !taginfo.isNormalTag()
         return
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -4046,12 +4046,17 @@ function! tagbar#jump() abort
     call s:JumpToTag(1)
 endfun
 
-" tagbar#jumpToNextTag() {{{2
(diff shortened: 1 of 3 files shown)


raven42:
Not providing a default keymap as this affects the actual file window and we don't want to change something the user may already have set. The function calls and new commands will be enough and the user can provide their own keymapping in their `.vimrc` to provide the correct mapping.

Recommended keymap:
```vim
nnoremap t[ :TagbarJumpPrev<CR>
nnoremap t] :TagbarJumpNext<CR>
```

PR #785, opened 2021-07-28
Fix for #784 - add additional option for `g:tagbar_wrap`

### LINKED ISSUE #784: Tagbar messes up identation on long function names and fields
When tagbar_wrap is set to 1 with linebreak, tagbar_wrap messes up indentation. A screenshot can be found [here](https://drive.google.com/file/d/111MQ7A6x2WInAmQ8FkDqE73FhmmTU10o/view?usp=sharing).

The only way to mitigate this issue is to set nolinebreak, however setting this from global init.vimrc is not a complete solution as people might want linebreak as well.

raven42:
Closes #784 

Adding additional setting to `g:tagbar_wrap` to allow for more granular
control of the `linebreak` functionality. If set to 1, linebreak will
also be set. If set to 2, linebreak will be disabled.

`g:tagbar_wrap` was added in #659

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1003,6 +1003,11 @@ function! s:InitWindow(autoclose) abort
         if exists('+linebreak')
             setlocal breakindent
             setlocal breakindentopt=shift:4
+            if g:tagbar_wrap == 1
+                setlocal linebreak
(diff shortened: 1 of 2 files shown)


daedroza (approved):
The feature works as intended. Great work, David!

alerque (approved):

PR #787, opened 2021-08-05
feature: Add basic Crystal support

### LINKED ISSUE #517: Add support for Crystal

dammer:
The Crystal [~/.ctags](https://github.com/dammer/crystal-ctags) file is need to work.

Peek 2021-08-05 22-05

Fixes #517

### DIFF
--- a/autoload/tagbar/types/ctags.vim
+++ b/autoload/tagbar/types/ctags.vim
@@ -195,6 +195,21 @@ function! tagbar#types#ctags#init(supported_types) abort
         \ {'short' : 's', 'long' : 'sections',          'fold' : 0, 'stl' : 1}
     \ ]
     let types.cobol = type_cobol
+    " Crystal {{{1
+    let type_crystal = tagbar#prototypes#typeinfo#new()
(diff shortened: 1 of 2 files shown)


alerque (approved):
This is looking alright to me.

By the way have you tried submitting the Crystal info upstream to Universal Ctags? They are usually pretty receptive to new language support.

raven42:
Looks good to me... only question would be if the `stl` flag should be 1 for all of them. Something like the macros for example I wouldn't think we'd want `stl=1`.

dammer:
Thanks Caleb
> By the way have you tried submitting the Crystal info upstream to Universal Ctags? They are usually pretty receptive to new language support.

Actually no, but I see requests for adding support. I will try to ask familiar C coders about write commit.

dammer:
@raven42 
> Looks good to me... only question would be if the `stl` flag should be 1 for all of them. Something like the macros for example I wouldn't think we'd want `stl=1`.

I checkd, in Crystal macro seems as defs not as in C macro.

raven42 (approved):

PR #788, opened 2021-08-23
Highlight key-map options in help file

BioBox:
It's easier to read this way. I glossed it over entirely the first time through.

### DIFF
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -437,60 +437,60 @@ KEY MAPPINGS                                                     *tagbar-keys*
 The following mappings are valid in the Tagbar window:
 
 <F1>/?        Display key mapping help.
-                Map option: tagbar_map_help
+                Map option: |tagbar_map_help|
 <CR>/<Enter>  Jump to the tag under the cursor. Doesn't work for pseudo-tags.
               If on generic header, it will fold/unfold that header.
-                Map option: tagbar_map_jump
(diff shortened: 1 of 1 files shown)


doc/tagbar.txt
```diff
@@ -437,60 +437,60 @@ KEY MAPPINGS                                                     *tagbar-keys*
 The following mappings are valid in the Tagbar window:
 
 <F1>/?        Display key mapping help.
-                Map option: tagbar_map_help
+                Map option: *tagbar_map_help*
```
raven42, line 440:
```suggestion
                Map option: |tagbar_map_help|
```
As per the `:help help-writing` documentation within vim, the `*tag*` syntax is used to define a tag, and the `|link|` is used to link to a tag. In this case we want to use the link to a tag option so please change all these to use the `|` characters instead.

raven42 (changes requested):

raven42:
Thanks for the contribution. Just need to change the format to a link format instead of a tag format and I'd be happy to help merge this in.

doc/tagbar.txt
BioBox: Okay, sure I'll run `%s/\*\(tagbar_map_\w\+\)\*/|\1|/g`.
It should be good now. This is my first time working on a help file btw.

raven42 (approved):

PR #792, opened 2021-10-02
Allow addition of new language types (#789)

### LINKED ISSUE #789: #777 breaks the ability to add a definition for a new language or file type
After merging #777, it is no longer possible to add a definition for a new language or file type as documented in `doc/tagbar.txt` and Vim produces the following error:

```
"sample.adoc" 41L, 2665B
Tagbar: Could not execute ctags for /home/jhradilek/sample.adoc!
Executed command: "'ctags' '--extra=' '--file-scope=yes' '-f' '-' '--format=2' '--excmd=pattern
' '--fields=nksSafet' '--sort=no' '--append=no' '--language-force=asciidoc' '--asciidoc-kinds=s
iIvaA' '--options=/home/jhradilek/.vim/pack/my-plugins/start/vim-asciidoc/ctags/asciidoc.cnf' '
/tmp/vMgKbzE/4.adoc'"
Command output:
ctags: Warning: Unsupported parameter 'e' for "fields" option
ctags: Unknown language "asciidoc" in "language-force" option
Exit code: 1
```

The problem is that tagbar now attempts to run `--language-force=asciidoc` before `asciidoc` is defined with the `--options` option:

```
$ ctags --language-force=asciidoc --options=asciidoc.cnf sample.adoc
ctags: Unknown language "asciidoc" in "language-force" option
```

Moving `--language-force=asciidoc` after `--options` resolves this problem:

```
$ ctags --options=asciidoc.cnf --language-force=asciidoc assembly_booting-a-beta-system-with-uefi-secure-boot.adoc
```

I am using Exuberant Ctags 5.8:

```
$ ctags --version
  Compiled: Jan 26 2021, 00:00:00
```

Moving the contents of the `asciidoc.cnf` file to `~/.ctags` of course solves this problem, but I would like to keep the configuration confined to my Vim plugin if at all possible.

jhradilek:
This is an attempt to resolve #789 and re-enable support for addition of a new language definition from within vim configuration without having to edit `~/.ctags`.

With this patch, if you need to define a language that is unknown to ctags and provide a custom file with `deffile` as documented in `doc/tagbar.txt`, you can use `deflang` as follows:

```vim
let g:tagbar_type_asciidoc = {
  \ 'ctagstype': 'asciidoc',
  \ 'deflang': 'asciidoc',
  \ 'deffile': expand('<sfile>:p:h:h') . '/ctags/asciidoc.cnf',
  \ 'sort': 0,
  \ 'kinds': [
    \ 's:Table of Contents',
    \ 'i:Included Files',
    \ 'I:Images',
    \ 'v:Videos',
    \ 'a:Set Attributes',
    \ 'A:Unset Attributes'
  \ ]
\}
```

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1397,6 +1397,12 @@ function! s:ExecuteCtagsOnFile(fname, realfname, typeinfo) abort
             let ctags_args += [ '-V' ]
         endif
 
+        " Define a new language before adding --language-force to the list of
+        " arguments
+        if has_key(a:typeinfo, 'deflang')
+            let ctags_args += ['--langdef=' . expand(a:typeinfo.deflang)]
+        endif
+
         " Third-party programs may not necessarily make use of this
         if has_key(a:typeinfo, 'ctagstype')
             let ctags_type = a:typeinfo.ctagstype


jhradilek:
On the off chance this solution gets approved, I'd be of course happy to update `docs/tagbar.txt` as well.

autoload/tagbar.vim
```diff
@@ -1397,6 +1397,12 @@ function! s:ExecuteCtagsOnFile(fname, realfname, typeinfo) abort
             let ctags_args += [ '-V' ]
         endif
 
+        " Define a new language before adding --language-force to the list of
+        " arguments
+        if has_key(a:typeinfo, 'deflang')
+            let ctags_args += ['--langdef=' . expand(a:typeinfo.deflang)]
+        endif
```
raven42, line 1404: We might be able to just key off the `deffile` instead. Then there wouldn't need to be any new config keys. Can you give this a try and see if this works?
```suggestion
        if has_key(a:typeinfo, 'deffile')
            let ctags_args += ['--langdef=' . a:typeinfo.ctagstype]
        endif
```

Issue #628, opened 2020-07-11
Tagbar on Windows issuing wrong command


afonsolage:
When I try to run `:TagbarToggle` I get the following error:
```
Tagbar: Ctags doesn't seem to be Exuberant Ctags!
BSD ctags will NOT WORK. Please [...]
Executed command: ""ctags" --version"
Command output:
ParserError:
Line |
   1 |  ""ctags" --version"
     |    ~~~~~~~~~~~~~~~~~
     | Unexpected token 'ctags" --version"' in expression or statement.
Exit code: 1
```
I'm using PowerShell Core and the command `""ctags" --version"` won't work because the extra quotes. So this is my shell config on NeoVim:
```
set shell=pwsh
set shellquote= shellpipe=\| shellxquote=
set shellcmdflag=-NoLogo\ -NoProfile\ -ExecutionPolicy\ RemoteSigned\ -Command
set shellredir=\|\ OutFile\ -Encoding\ UTF8
```

It seems that Tagbar is ignoring the shellquote and shellxquote configurations.

If I run `ctags --version` I got the following output:
```
Universal Ctags 0.0.0(79799bd4), Copyright (C) 2015 Universal Ctags Team                                                                                                                       
Universal Ctags is derived from Exuberant Ctags.                                                                                                                                               
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert                                                                                                                                    
  Compiled: Jul 10 2020, 00:08:01                                                                                                                                                              
  URL: https://ctags.io/                                                                                                                                                                       
  Optional compiled features: +win32, +wildcards, +regex, +internal-sort, +unix-path-separator, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc                                                                                                                                                                                           
```

There is any workaround for this?

Thanks!

alerque:
I don't think we have any support for `pwsh` yet. Windows is supported other ways, but not with the VIM shell changed like this.

I'd be happy to accept a contribution that fixed this for first class Powershell support and escaping everything right across platforms, but I don't have access to or experience with Windows to make this happen myself. If you'd like to take a stab at it go ahead. I think there are some other shell support functions you could model from.

avysk:
I have very much the same problem.

My settings are
```
  set shell=pwsh\ -nop\ -nol
  set shellcmdflag=-c
```

Commenting these out makes TagBar to run.

avysk:
[...]

Issue #424, opened 2017-07-12
Windows Powershell


supermyz:
hope this can help someone
If you set vim default shell set to 'powershell' on Windows 10, Tagbar won't recognize ctags output and hence won't work as expected

![image](https://user-images.githubusercontent.com/27808031/28118853-116e26ec-6713-11e7-8fa9-fb4d1082b5ee.png)

mqudsi:
That doesn't seem to be a tagbar issue, does `:tag tag_name` work?

alerque:
I do not have a Windows machine for testing, but I'd be happy to review a PR that fixes this issue, especially if other people can jump in to confirm what is going on here. That being said I see there was a request for information already @supermyz, did you have any feedback on this?

Is anybody else able to check and confirm or deny that this is an issue still?

PR #798, opened 2021-11-26
Add support for powershell

fwy:
I use tagbar on Windows 10, and made a small change to allow tagbar to work when powershell is being used as the vim shell.

Before this change, when opening a tagbar window I receive the misleading message "Ctags doesn't seem to be Exuberant Ctags!" that was reported in issues https://github.com/preservim/tagbar/issues/424 and https://github.com/preservim/tagbar/issues/628.

After the change, I am able to use tagbar with either the original Windows powershell (powershell.exe) or the newer open-source powershell (pwsh.exe).

The change works in both neovim and vim, but I found a posting (https://groups.google.com/g/vim_dev/c/Hcega_rfWOQ?pli=1) that highlights a separate problem using powershell in vim.  In vim, the &shellquote variable is automatically set to double-quote (")  when running on Windows whenever &shell contains the string 'sh' (mentioned in the vim powershell help), which breaks powershell commands.  To work around the issue, I set &shell to 'pwSh.exe' with an uppercase 'S' in my _vimrc file.  In neovim, this problem does not occur.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2992,7 +2992,7 @@ function! s:EscapeCtagsCmd(ctags_bin, args, ...) abort
 
     "Set up 0th argument of ctags_cmd
     "a:ctags_bin may have special characters that require escaping.
-    if &shell =~? 'cmd\.exe$' && a:ctags_bin !~# '\s'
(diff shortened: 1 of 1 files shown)


alerque (approved):
Thanks for the contribution. I don't have any Windows machines to test on, but the code logic looks sane to me. I'll be happy to adjust if anybody reports problems, but this seems like a good step. Thanks for taking the time to help others.

PR #800, opened 2021-12-08
Add support of TagbarOpen arguments for TagbarToggle

whuji:
I needed TagbarToggle to pass arguments to TagbarOpen. It was already implemented in TagbarToggle() but not in the command. So it is.

### DIFF
--- a/plugin/tagbar.vim
+++ b/plugin/tagbar.vim
@@ -183,8 +183,8 @@ augroup TagbarSession
 augroup END
 
 " Commands {{{1
-command! -nargs=0 Tagbar              call tagbar#ToggleWindow()
-command! -nargs=0 TagbarToggle        call tagbar#ToggleWindow()
+command! -nargs=? Tagbar              call tagbar#ToggleWindow(<f-args>)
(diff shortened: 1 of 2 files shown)


plugin/tagbar.vim
```diff
@@ -184,7 +184,7 @@ augroup END
 
 " Commands {{{1
 command! -nargs=0 Tagbar              call tagbar#ToggleWindow()
```
raven42, line 186:
```suggestion
command! -nargs=? Tagbar              call tagbar#ToggleWindow(<f-args>)
```

raven42 (changes requested):
We should be consistent with the behavior of both `:TagbarToggle` and `:Tagbar`.

Also it would be ideal if you could update the documentation for these commands in the `doc/tagbar.txt` file to indicate the optional arguments and state they are the same arguments that are supported for the `:TagbarOpen` command.

whuji:
I have added the same behavior for :Tagbar and a line in the doc to specify the arg.

raven42 (approved):
Looks good. Thanks for the contribution!

PR #805, opened 2022-01-20
Only set ballooneval while in tagbar window

### LINKED ISSUE #499: Tagbar sets ballooneval globally on init
Hi,

Thanks for Tagbar, I find it really useful :-)

I've added some horrible little hacks to my vimrc to deal with the fact that it sets ballooneval globally when setting up the window though (i.e. `set ballooneval` rather than `setlocal ballooneval`).

Is there a reason why ballooneval is set using `set` rather than `setlocal`?

Thanks

Mark

mmrwoods:
ballooneval is a global option, so setting it while initialising tagbar
permanently enables ballooneval functionality for the GUI, overriding
the vim default (disabled) or a user explicitly setting noballooneval.

To address this, only set ballooneval when entering the tagbar window,
and restore the original ballooneval setting when leaving the window.

This is not perfect, but there is precedent, Netrw behaves like this.

Fixes #499

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -563,6 +563,16 @@ function! s:CreateAutocommands() abort
             autocmd WinEnter   __Tagbar__.* call s:SetStatusLine()
             autocmd WinLeave   __Tagbar__.* call s:SetStatusLine()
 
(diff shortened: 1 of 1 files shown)


raven42 (approved):
Looks good to me. Thanks for the contribution.

PR #807, opened 2022-02-05
Use tagbar_iconchars not present in emoji

dexgs:
The previous tagbar_iconchars, specifically "▶" (U+25B6), has a representation in emoji which can display strangely in some terminal emulators. Simply using triangle characters that don't overlap with emoji resolves the issue.


alerque (approved):
Thanks for taking the time to contribute. This seems reasonable.

PR #810, opened 2022-03-09
Don't show errors when opening preview window 

zharmany:
I was experiencing an issue where, if the current buffer had unsaved changes, the action of opening the preview window on a tagbar tag would cause an error that would require a keypress to dismiss. I found this jarring and simply updated the `pedit` command to fail silently.

I'm hoping that someone else finds this useful, or perhaps there's a better solution than just ignoring the error.

Cheers!

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2500,7 +2500,7 @@ function! s:ShowInPreviewWin() abort
     " Open the preview window if it is not already open. This has to be done
     " explicitly before the :psearch below to better control its positioning.
     if !pwin_open
-        silent execute
+        silent! execute
             \ g:tagbar_previewwin_pos . ' pedit ' .
             \ fnameescape(taginfo.fileinfo.fpath)
         if g:tagbar_position !~# 'vertical'
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2500,9 +2500,12 @@ function! s:ShowInPreviewWin() abort
     " Open the preview window if it is not already open. This has to be done
     " explicitly before the :psearch below to better control its positioning.
     if !pwin_open
+        let l:confirm = &confirm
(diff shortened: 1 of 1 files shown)


raven42:
Can you describe an example of this error a little more? based on what you described, I tried the same thing and I don't see any error opening the preview window. To attempt to reproduce, I had a file opened with the tagbar window opened and fully populated, then I edited the file to add some innocuous code, then I went to the tagbar window and hit `<SHIFT-P>` on another tag either before or after the new code, and the preview window opened fine and showed the correct preview.

Does it need to be an entirely new file? Are you able to share a screenshot of the error you are seeing?

zharmany:
Hey there, thanks for the reply.

I made a quick screen recording to show what I mean. It could be the behavior of automatically opening the preview window which uses the `pedit` function. I'm still very much learning vimscript.

https://user-images.githubusercontent.com/1714524/158079674-58a6c2da-969f-4f9a-b4b9-c70c910c407e.mov

raven42:
Ok.... I am able to reproduce now. Turns out if you have `set confirm` in your .vimrc, it changes the behavior of this. So we may actually want to incorporate that into this as well.

raven42 (changes requested):
This appears to be an issue with vim when opening a modified buffer via `pedit <file>`. You are correct that with the `silent! ...` it does fix this, but then there is also an issue if the user has `set confirm` in their .vimrc. This addition will cache the current value, then unset it, and finally restore it once the preview window is open. This will ensure consistent behavior with what we want for the preview window.

zharmany:
Thanks again for looking at it, it looks like your proposed change would be more appropriate!

raven42 (approved):
This looks good to me. 

@alerque, do you have any issues with this going in?

alerque (approved):

Issue #799, opened 2021-12-04
Wrong tags with javascript specific syntax


igoracmelo:
Sorry if it is not tagbar's "fault", but when using this type of syntax with javascript the tags aren't quite right...
![image](https://user-images.githubusercontent.com/85039990/144715792-84d66bf3-30e0-403b-8f63-197df8edaf15.png)

This works:
![image](https://user-images.githubusercontent.com/85039990/144715886-79223b04-ccdc-4c43-815e-e56bdd5a5f4e.png)

raven42:
This would need to be brought up with the universal-ctags team first. I tested the code you have and currently ctags doesn't recognize the `async <name> ()` format as a function. Here is the raw ctags output that tagbar has to work with. Once universal ctags can support this function syntax, we can verify that tagbar can detect and display it correctly.

```
tagbar (master) tagbar-test-files$ ctags --extras=+F -f - --format=2 --excmd=pattern --fields=nksSafet --sort=no --append=no -V --language-force=javascript --javascript-kinds=vcCgpmf test.js
  Option: --language-force=javascript
  Option: --javascript-kinds=vcCgpmf
Initialize parser: JavaScript
Reading command line arguments
OPENING test.js as JavaScript language file [new]
response	test.js	/^		const response = await axios.get('api\/school-system\/users\/me')$/;"	C	line:5
formData	test.js	/^		const formData = new FormData()$/;"	C	line:10
response	test.js	/^		const response = await axiot.put('api\/school-system\/users\/me', payload)$/;"	C	line:20
response	test.js	/^		const response = await axios.put('api\/auth\/change-password', user)$/;"	C	line:25
response	test.js	/^		const response = await axios.post('api\/auth\/password-recovery\/change-password', user)$/;"	C	line:31
tagbar (master) tagbar-test-files$
```

raven42:
@igoracmelo I've pushed up a change that might give partial support for this. Take a look at #811 and see if this will work for your case. I tested it out using the example you gave and the following definitions in my .vimrc and it seemed to work for me.

```vim
	let g:tagbar_type_javascript = {
				\ 'ctagstype'	: 'javascript',
				\ 'regex'		: [
					\ '/async[ \t]+(.*)[ \t]*\(.*\)/\1/f/func/function/',
				\ ],
				\ 'kinds'		: [
					\ 'v:global variables:0:0',
					\ 'C:constants:0:0',
					\ 'c:classes:0:1',
					\ 'g:generators:0:0',
[...]

PR #812, opened 2022-03-24
Fixes #811, #799: Add language specific regex support for custom tag kinds

### LINKED ISSUE #811: Add global support for TODO FIXME XXX...
So tagbar is great for navigating projects. But I feel like implementing a way to navigate TODO's (and all the others) would improve it greatly.  I was looking around for a solution and stumbled upon [this ctags flag][1] which seemed promising but the given example is exclusive to javascript.

So what I think could be great is to take a page from their book but make it universal. Therefore this:

`--regex-javascript=/\/\/[ \t]*\(FIXME\)[ \t]*:?.*/\1/T,Todo,Todo messages/b`

which takes JS specific comments  into account could probably be turned into something like this:

`--regex-all=\(FIXME\)[ \t]*:?.*/\1/T,Todo,Todo messages/b`

which is language agnostic since comments simply aren't handled (I think closing comments would be kept i.e. `*/` but it's better than nothing).

Please note that I haven't tested any of these commands. I assumed `all` could be used as a valid language since the ctags manual mentions that `--alias-all=...` can be used for the `--alias-<LANG>=...` command. I simply thought this could be a great starting point if such a feature were to be implemented into vim-tagbar. I don't know if this should be added as a default. Perhaps it should be implemented as an option like:

`let g:tagbar_todo_enable=1`

Anyway, have a great day.

[1]:https://gist.github.com/romainl/27c1c29462a8c6c868f1a9bf244bc71d

raven42:
Fixes #811 and #799

This will add custom tag kind support on a per-language level. It is not feasible to provide a global definition as the tag kind identifier must be unique, and each language has a different set of kinds defined for it. So this way we can at least allow a per-language custom definition for any custom tags that the user may wish to add using a `--regex-<lang>=<regex>` option for the ctags command. This will be flexible enough to allow adding the `TODO` type tag kinds as requested in #811, but will also provide a framework for a more general approach to allow any custom tag kinds to be added based on what the user is needing.

For the case in #799, we can add the following regex definition. Note: this does not allow for scope information for the function, but it does at least provide an option for the `async <function> (<args>)` type syntax which is better than what we have today.
```vim
let g:tagbar_type_javascript = {
        \ 'ctagstype'	: 'javascript',
        \ 'regex'		: [
            \ '/async[ \t]+(.*)[ \t]*\(.*\)/\1/f/func/function/',
        \ ],
        ... # Rest of the definition here
    \ }
```

### DIFF
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -1664,7 +1664,7 @@ ctagsargs: The arguments to be passed to the filetype-specific ctags program
                        \ 's:structs:0:1',
                        \ 'm:members:1:0',
                        \ 'v:variables:0:0',
-                       \ 'f:functions:0:1:
+                       \ 'f:functions:0:1',
                        \ 'T:todo:0:0',
                    \ ],
                    \ 'sro'          : '::',
--- a/doc/tagbar.txt
+++ b/doc/tagbar.txt
@@ -1637,6 +1637,47 @@ ctagsargs: The arguments to be passed to the filetype-specific ctags program
 
            If special escaping is required for different OS shell types or if
            in doubt, then it is recommended to define ctagsargs with a List.
+ regex:    A list of regular expression arguments that will be provided to
+           ctags via a '--regex-<lang>=<regex>' option. Multiple regular
+           expressions can be provded to allow adding support for multiple new
+           tag regular expressions. The regex statement should be formatted as
+           documented in the ctags documentation.
+            https://docs.ctags.io/en/latest/optlib.html#regex-option-argument-flags
+           When adding a regex statement, the corresponding tag kind label
+           will need to be defined in the |kinds| definition in the structure
+           as well. An example for a 'TODO' label for the c language would be
+           as follows (note: use of the '{_anonymous=..}' option is not
+           strictly necessary, but does give a unique tag name for each match):
+>
+               let g:tagbar_type_c = {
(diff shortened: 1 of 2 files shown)


doc/tagbar.txt
```diff
@@ -1637,6 +1637,47 @@ ctagsargs: The arguments to be passed to the filetype-specific ctags program
...
+                       \ 'h:header files:1:0',
+                       \ 'd:macros:1:0',
+                       \ 'p:prototypes:1:0',
+                       \ 'g:enums:0:1',
+                       \ 'e:enumerators:0:0',
+                       \ 't:typedefs:0:0',
+                       \ 's:structs:0:1',
+                       \ 'm:members:1:0',
+                       \ 'v:variables:0:0',
+                       \ 'f:functions:0:1:
```
ChausseBenjamin, line 1667: There's seem to be a typo... the string for f:functions is not closed (and there is no comma at the end).
The line should be the following I think:
```
       \ 'f:functions:0:1:',
```

raven42: ah good catch... that's what i get for copy/paste into the doc itself instead of actually loading it up... oops. Thanks for catching that. I'll get it fixed and push it up.

alerque (approved):
I was about to say this shouldn't have gotten past our CI ... then I saw it was in docs. oops ;-)

PR #820, opened 2022-06-09
Allow autocmds in the host window when s:JumpToTag() leaves the tagbar

lyokha:
This fixes cursor splashing in wrong positions when plugin [Beacon](https://github.com/DanilaMihailov/beacon.nvim) is enabled.

Recently I installed the Beacon plugin and found that cursor splashes it runs on `WinEnter` after the cursor jumps from the tagbar to the host window happen in wrong places (the top or the bottom line of the host window). The following animation shows what happens:

Peek 2022-06-08 23-51

Note that splashes occur at the top and the bottom lines of the host window when the tag gets chosen in the tagbar and the cursor jumps back to the host window.

The reason is that `s:JumpToTag()` makes jump with `noautocmd` on.

Look what it does (this is the original, not yet changed code):

```vim
    if a:stay_in_tagbar
        call s:HighlightTag(0)
        call s:goto_win(tagbarwinnr)
        redraw
    elseif g:tagbar_autoclose || autoclose
        " Also closes preview window
        call s:CloseWindow()
    else
        " Close the preview window if it was opened by us
        if s:pwin_by_tagbar
            pclose
        endif
        if s:is_maximized
            call s:ZoomWindow()
        endif
        call s:HighlightTag(0)
    endif
```

There are 3 cases:

1. Stay in the tagbar after jump -> call `s:HighlightTag()` -> `noautocmd` in the host window, that's Ok as we are not moving from the tagbar.
2. Autoclose tagbar -> autocmds are on as `s:CloseWindow()` runs them, that's Ok too.
3. Jump from the tagbar window to the host window -> call `s:HighlightTag()` -> `noautocmd` in the host window, that seems to be Not Ok as the cursor physically moves to the host window.

Function `s:HighlightTag()` calls `s:goto_win()` with `noautocmd` flag. In this PR I extended trailing arguments of `s:HighlightTag()` to let `s:JumpToTag()` call this with the unset `noautocmd` flag passed to `s:goto_win()` for the host window in the case 3.

After the fix, the Beacon cursor splashes occur in the right positions.

Peek 2022-06-08 23-49

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -883,7 +883,7 @@ function! s:OpenWindow(flags) abort
     if tagbarwinnr != -1
         if winnr() != tagbarwinnr && jump
             call s:goto_win(tagbarwinnr)
-            call s:HighlightTag(g:tagbar_autoshowtag != 2, 1, curline)
+            call s:HighlightTag(g:tagbar_autoshowtag != 2, 1, 1, curline)
         endif
         call tagbar#debug#log('OpenWindow finished, Tagbar already open')
         return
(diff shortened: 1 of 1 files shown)


raven42:
changes look good to me. Thanks for the contribution.

PR #822, opened 2022-07-14
Add `autoclose_netrw` configuration option

### LINKED ISSUE #821: Close pane with tagbar would also close the nerdtree pane
When I open a file from nerdtree in split view,(press i), and toggle tagbar for the file, then close the file, the nerdtree pane would also be closed. How can I disable this behaviour?

raven42:
Closes #821

This will add the `autoclose_netrw` configuration option. If set to 1
this will close out of vim if tagbar and netrw (or nerdtree) are the
only remaining windows. If set to 0, then this will leave the netrw (or
nerdtree) window if that is the only remaining window once tagbar is
closed.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3509,7 +3509,15 @@ function! s:HandleOnlyWindow() abort
     let vim_quitting = s:vim_quitting
     let s:vim_quitting = 0
 
-    if vim_quitting && !s:HasOpenFileWindows()
+    let file_open = s:HasOpenFileWindows()
+
+    if vim_quitting && file_open == 2 && !g:tagbar_autoclose_netrw
+        call tagbar#debug#log('Closing Tagbar due to QuitPre - netrw only remaining window')
+        call s:CloseWindow()
+        return
+    endif
+
+    if vim_quitting && file_open != 1
         call tagbar#debug#log('Closing Tagbar window due to QuitPre event')
         if winnr('$') >= 1
             call s:goto_win(tagbarwinnr, 1)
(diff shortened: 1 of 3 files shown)


alerque (approved):

raven42:
@alerque I was wondering your opinion on a point in this PR.

If I'm understanding this correctly, in the current code, the `s:HasOpenFileWindows()` function would return a TRUE value if the netrw window was open. So in the event someone does not have nerdtree installed, and they follow the steps in issue #821, then I believe the netrw window would remain open.

Now if they do have nerdtree installed, as nerdtree provides an override for the netrw window, then if they follow the procedure in the issue, then as @WeissZucker pointed out, then the nerdtree window closes.

So onto this PR... one of the goals of this is to provide consistent behavior for both the netrw and nerdtree windows, but at the same time I don't want to disrupt the workflow of people that are used to the current behavior. With current behavior, they each behave differently. So now adding the configuration option does allow for customization, however I want to make sure we set the default value correctly. I'm making an assumption that most people that are using tagbar, probably are also using nerdtree, so I set the default for the `g:tagbar_autoclose_netrw` to TRUE, so people that do have nerdtree will have the same workflow. However people that have tagbar, but not nerdtree will see a change in behavior. Alternatively we could swap the default value, but then we'd have a change in behavior for people that do have both tagbar and nerdtree installed.... So I'm curious how best to handle this, and wanted to get your thoughts on it.

alvan:
Do you have any plans to merge this pull request? It solves a very serious bug I encountered, that is, when using bufexplorer, nerdtree, and tagbar at the same time, operations in bufexplorer (such as deleting a buffer) have unexpected errors.

raven42:
@alvan yes I can merge it in. I was waiting for feedback from the community or other moderators as to the default value, however I can merge it in as is and then we can swap the default later if the default becomes an issue.

Shane-XB-Qian:
i wish this default value is to `0`
as common sense, any default setting should not broken the consistent/previous behavior.
// wish we can keep it while in maintain sth, meant `compatible`.
at least, some user used `netrw` as 'project viewer', then had high possibility that's the last window!

alvan:
@raven42 
Thank you so much for the update! After the update it worked very well in my use case.
And I also set: `let g:tagbar_autoclose_netrw = 0` to minimize interference between plugins.

raven42:
@Shane-XB-Qian / @alvan, I've changed the default to 0 based on your feedback. It seems more people are in favor of having this disabled by default, so we'll go with that.

Thanks for the feedback.

PR #842, opened 2022-11-29
Replace vim-easytags link with vim-gutentags in README

jdholtz:
Since [vim-easytags](https://github.com/xolox/vim-easytags) has not been updated in seven years, it would be best to point users to [vim-gutentags](https://github.com/ludovicchabant/vim-gutentags) for a vim plugin that manages your tags automatically, as that is being actively maintained.


raven42:
This looks ok to me. I don't use either of these plugins though.

@alerque do you have any input on this before I merge it in?

alerque (approved):
I'm also a gutentags user and this change seems smart to me; if we're pointing out some random example of a plugin for a different purpose than the current one we might as well point them at something maintained and that we actually use.

PR #853, opened 2023-04-08
Fixed single-click

blockchainian:
this pr fixes left click so that it opens/closes the folding.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -2945,9 +2945,9 @@ function! s:CheckMouseClick() abort
     let line   = getline('.')
     let curcol = col('.')
 
-    if (match(line, g:tagbar#icon_open . '[-+ ]') + 1) == curcol
+    if match(line, g:tagbar#icon_open) == 0
         call s:CloseFold()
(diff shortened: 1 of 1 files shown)


raven42:
I believe the original intent of the feature was only to allow folding when the open/close character is clicked. It looks like with this change, that is being expanded to open/close the fold if the click occurs anywhere on the line. Am I interpreting that correctly? I think I'm ok with that change, but it would need to be updated to account for nested folds.

If we take a python script like this:
```python
def my_func():
    print("first layer")

    def inner_func():
        print("second layer")

        def inner_inner_func():
            print("third layer")
        print("went back to the second layer, but tag points to third layer")
    print("went back to the first layer, but tag points to third layer")
```

Then the fold open/close doesn't work for the internal definition of `inner_func()`.

If you can make it work to so this type of definition works as well, then I would be ok merging in this change.

raven42 (changes requested):
This breaks nested folded functionality.

blockchainian:
closing this pull request since i have switched to nvim tree.

PR #857, opened 2023-08-14
Add native bibtex support for #856

### LINKED ISSUE #856: Error on configuration of g:tagbar_type_bib
I think there is an error on the suggested configuration for Bibtex files on the [ Wiki page](https://github.com/preservim/tagbar/wiki#bib).

The `ctagstype` variable must be set to `bibtex`, instead of `bib`. When i use `bib`, it does not work.

By the way, `Misc` references are not shown on Tagbar, although other types of references (Article, Books, etc) work.

raven42:
Fixes #856

TODO:
- [ ]: Need to identify which of the `kinds` are foldable
- [ ]: Need to identify which of the `kinds` are scoped (stl)
- [ ]: Need to validate all the different kinds.

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -187,6 +187,28 @@ function! tagbar#types#uctags#init(supported_types) abort
         \ {'short' : 'v', 'long' : 'patterns',  'fold' : 0, 'stl' : 1}
     \ ]
     let types.beta = type_beta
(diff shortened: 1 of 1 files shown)


raven42:
Tested with user test file and validated proper functionality. ctags doesn't give any scope information, and user confirmed all tags are global scope, so no stl / fold flags are needed.

alerque (approved):

PR #858, opened 2023-08-14
#682 gvim & pythonx

### LINKED ISSUE #682: Tagbar doesn't work on gvim for windows with python installed
Hi,
When I run TagbarToggle I get the following error:

	OSError: [WinError 6] The handle is invalid

But if debug mode is enabled, Tagbar work fine. In debug mode, the plugin does not use python to call ctags:
```vim
    if tagbar#debug#enabled()
        silent 5verbose let ctags_output = system(a:ctags_cmd)
        call tagbar#debug#log(v:statusmsg)
        call tagbar#debug#log('Exit code: ' . v:shell_error)
        redraw!
    else
        let py_version = get(g:, 'tagbar_python', 1)
        silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
    endif
```
The plugin uses undocumented parameter g:tagbar_python, which allows using python for ctags calls.
By default, this parameter set to 1, which launches call via python by default.
This parameter is processed here:
```vim
    if has('win32') && !has('nvim') && a:version > 0 && (has('python3') || has('python2'))
        if a:version == 3 && has('python3')
            let pyx = 'py3 '
            let python_eval = 'py3eval'
        elseif a:version == 2 && has('python2')
            let pyx = 'py2 '
            let python_eval = 'pyeval'
        else
            let pyx = 'pyx '
            let python_eval = 'pyxeval'
        endif
```
Next, the part of the code where the ctags call is formed via Python:
```vim
        exec pyx . 'import subprocess, vim'
        exec pyx . '__pp = subprocess.Popen(**__argv)'
        exec pyx . '__return_text = __pp.stdout.read()'
        exec pyx . '__pp.stdout.close()'
        exec pyx . '__return_code = __pp.wait()'
        exec 'let l:hr = '. python_eval .'("__return_text")'
        exec 'let l:pc = '. python_eval .'("__return_code")'
```
Calling subprocess from gvim does not work, if "stdin" not defined.
Bellow are the patches to fix this behaviour:

```patch
--- "a\\autoload\\tagbar.vim"	2020-10-24 00:17:10.000000000 +0300
+++ "b\\autoload\\tagbar.vim"	2020-10-24 21:29:14.307119000 +0300
@@ -2917,6 +2917,7 @@
         let l:pc = 0
         exec pyx . 'import subprocess, vim'
+        exec pyx . '__argv["stdin"] = subprocess.DEVNULL'
         exec pyx . '__pp = subprocess.Popen(**__argv)'
@@ -2971,7 +2972,7 @@
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 1)
+        let py_version = get(g:, 'tagbar_python', 0)
         silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
     endif
 

```

fdobad:
Fixes #682

Thanks to [emclimber](https://github.com/emclimber)

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3163,7 +3163,11 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 1)
+        if has('win32')
+          let py_version = get(g:, 'tagbar_python', 0)
+        else
+          let py_version = get(g:, 'tagbar_python', 1)
+        endif
         silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
     endif
 
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3163,7 +3163,7 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 0)
+        let py_version = get(g:, 'tagbar_python', 1)
         silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
     endif
 
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3107,6 +3107,7 @@ function! s:run_system(cmd, version) abort
         let l:pc = 0
         exec pyx . 'import subprocess, vim'
         exec pyx . '__argv = {"args":vim.eval("a:cmd"), "shell":True}'
+        exec pyx . '__argv["stdin"] = subprocess.DEVNULL'
         exec pyx . '__argv["stdout"] = subprocess.PIPE'
         exec pyx . '__argv["stderr"] = subprocess.STDOUT'
         exec pyx . '__argv["errors"] = "ignore"'
@@ -3162,7 +3163,7 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 1)
+        let py_version = get(g:, 'tagbar_python', 0)
         silent let ctags_output = s:run_system(a:ctags_cmd, py_version)
     endif
 
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ If the ctags executable is not installed in one of the directories in your
 `$PATH` environment variable you have to set the `g:tagbar_ctags_bin`
 variable, see the documentation for more info.
 
+Windows users should add `let g:tagbar_python = 0` to their vimrc.
+
 ## Quickstart
 
 Put something like the following into your ~/.vimrc:
--- a/README.md
+++ b/README.md
@@ -46,8 +46,6 @@ If the ctags executable is not installed in one of the directories in your
 `$PATH` environment variable you have to set the `g:tagbar_ctags_bin`
 variable, see the documentation for more info.
 
-Windows users should add `let g:tagbar_python = 0` to their vimrc.
-
 ## Quickstart
 
 Put something like the following into your ~/.vimrc:


alerque (approved):
This seems to make sense to me, but not being on Windows I'm not qualified to test it. It would be helpful if the original reporter or anybody else with the issue could chime in to mention if this works for them. Baring any feedback we can probably just merge eventually and see if anybody complains, but it would be nice to have at least one person verify it first.

autoload/tagbar.vim
```diff
@@ -3162,7 +3163,7 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 1)
+        let py_version = get(g:, 'tagbar_python', 0)
```
raven42, line 3166: I would prefer we don't change the default behavior for the `g:tagbar_python` variable. Initializing the stdin to DEVNULL is fine, but can we verify if setting the default for `g:tagbar_python = 0` is really necessary? Would it be possible to have this as a user setting so we aren't changing the default? We can add documentation around this variable to indicate when it should be set to 0.

fdobad: Unluckily `g:tagbar_python = 1` reverts to the same problem.  

Tried it with gvim9.0.1672_64 on windows 11 (`> winget install --id vim.vim`), with python3.11.4. [ctags5.8](https://sourceforge.net/projects/ctags)

image

raven42: You are still seeing this after you set the `let g:tagbar_python = 0` in your vimrc? That should set it to 0 for you and it won't use the default.

fdobad: That works! 
Guess a install for windows users hint should be in place? Like

Windows users should add `let g:tagbar_python = 0` to their `vimrc`

alerque: I am not following closely enough to suggest an exact solution here, but somehow or another if we're at a point of recommend all Windows users do something special we should just make that the default for Windows. Or whatever the situation, the out of box configuration for a given platform should be the best defaults for that platform without most people needing to set something special.

raven42:
@fdobad can you try this change and see if this addresses it too? This could be a better solution for everyone.

https://github.com/preservim/tagbar/commit/d73de9c3a7100e886e6a83cbc429b37bb7161b5f

autoload/tagbar.vim
```diff
@@ -3162,7 +3163,11 @@ function! s:ExecuteCtags(ctags_cmd) abort
         call tagbar#debug#log('Exit code: ' . v:shell_error)
         redraw!
     else
-        let py_version = get(g:, 'tagbar_python', 1)
+        if has('win32')
+          let py_version = get(g:, 'tagbar_python', 0)
+        else
+          let py_version = get(g:, 'tagbar_python', 1)
+        endif
```
raven42, line 3170: Unfortunately this change doesn't make a whole lot of sense. Inside the `s:run_system()` function call, there is a check already for `if has('win32')` and is specifically going into the python handler for that. The original issue was raised specifically against gvim, not just vim. So we need to understand if this is required for all windows installs, or if this is specific to only gvim installations on windows. If we can get some confirmation if this is only for gvim, then we would want to handle this in the `s:run_system()` function adding a `!has('gui_running')` check:

```vim
function! s:run_system(cmd, version) abort
    if has('win32') && !has('nvim') && !has('gui_running') && a:version > 0 && (has('python3') || has('python2'))
```

Otherwise if this does effect every windows installation, then still it would be better to handle it in the `s:run_system()` function call adding a `!` in front of the `has('win32')` check:
```vim
function! s:run_system(cmd, version) abort
    if !has('win32') && !has('nvim') && a:version > 0 && (has('python3') || has('python2'))
```

So if someone can help confirm the correct behavior on windows for both GUI and non-GUI installations, then we should update it like one of these options.

fdobad:
tag_bar_works_3ways

I'm so sorry, now I'm trying to figure out which plugin was messing it up.

PR #861, opened 2023-09-21
Add support for biblatex

### LINKED ISSUE #856: Error on configuration of g:tagbar_type_bib

raven42:
Fixes #856

Note: VIM does not have a known file extension to differentiate between `.bib` and a BibLaTeX file. From what I can see, there is no common file-extension for a BibLaTeX file, so a custom filetype configuration will be needed in vim to differentiate between the two.

### DIFF
--- a/autoload/tagbar/types/uctags.vim
+++ b/autoload/tagbar/types/uctags.vim
@@ -209,6 +209,49 @@ function! tagbar#types#uctags#init(supported_types) abort
     \ ]
     let types.bibtex = type_bibtex
     let types.bib = type_bibtex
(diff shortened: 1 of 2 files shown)


alerque (approved):

raven42:
image

PR #866, opened 2024-01-13
fix: 'g:tagbar_autoclose_netrw' was not actually really default to 0

Shane-XB-Qian:
https://github.com/preservim/tagbar/pull/822
it seems did not actually really reset the default value to 0, 
or just the init and doc, but the logic still was 1.
// and the logic seems was not correct, try to fix.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -3521,13 +3521,15 @@ function! s:HandleOnlyWindow() abort
 
     let file_open = s:HasOpenFileWindows()
 
-    if vim_quitting && file_open == 2 && !g:tagbar_autoclose_netrw
-        call tagbar#debug#log('Closing Tagbar due to QuitPre - netrw only remaining window')
-        call s:CloseWindow()
+    if vim_quitting && file_open == 2
+        if g:tagbar_autoclose_netrw
+            call tagbar#debug#log('Closing Tagbar due to QuitPre - netrw only remaining window')
+            call s:CloseWindow()
+        endif
         return
     endif
 
-    if vim_quitting && file_open != 1
+    if vim_quitting && file_open == 0
         call tagbar#debug#log('Closing Tagbar window due to QuitPre event')
         if winnr('$') >= 1
             call s:goto_win(tagbarwinnr, 1)


raven42:
I did a quick test with this change, and it did not work as expected for me. I had the following windows open:
* File Window
* Tagbar Window
* NerdTree Window

Then in my config file I have
```
let g:tagbar_autoclose_netrw = 1
```

With this my expected behavior is that when I type a `:q` on the file window, all three windows close. With the current code on perservim:master I see this behavior. When I checked out to your branch, I no longer see this behavior. When I type `:q` on your branch, I see the NerdTree window open still, but tagbar and the file window close.

Can you please describe the reasoning behind this change? What is the behavior you are expecting vs what is the behavior you are seeing?

Shane-XB-Qian:
> let g:tagbar_autoclose_netrw = 1


> With this my expected behavior is that when I type a :q on the file window, all three windows close. With the current code on perservim:master I see this behavior.


what's vim version you are using?

when it is `= 1`, even with my PR, actually it would be a err saying: windows layout cannot be changed in autocmd, when it is doing 'close'.

though that should be another story.


> When I checked out to your branch, I no longer see this behavior. When I type :q on your branch, I see the NerdTree window open still, but tagbar and the file window close.


when it's `= 1`, and No this PR, same it would be an err like above, but when it is doing 'bdelete'.


--
shane.xb.qian

raven42:
Ah ok, wonder if it is something with the vim version.

I'm currently using vim 8.2 still. I've played around a little bit with vim 9.0, but that isn't my main yet as I've seen some issues with plugins and such.

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 11 2023 23:53:12)
Included patches: 1-3995, 4563, 4646, 4774, 4895, 4899, 4901, 4919
```

Shane-XB-Qian:
> Ah ok, wonder if it is something with the vim version.
> 
> I'm currently using vim 8.2 still. I've played around a little bit with vim 9.0, but that isn't my main yet as I've seen some issues with plugins and such.
> 
> ```
> VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 11 2023 23:53:12)
> Included patches: 1-3995, 4563, 4646, 4774, 4895, 4899, 4901, 4919
> ```

yes, let's hold on this PR a while, let's see if the vim itself would make plugins life easy. :shrug:
// compared to 'err', i am more Not like 'crash', so let's hold.

-- 
shane.xb.qian

Shane-XB-Qian:
upstream ticket in vim repo had been closed, and recently there are some vim window operation code/logic were patched, so supposed now it was ok? or whatever, please expressed your thinking there about if vim itself can make plugin's life easy.. :smile:

Shane-XB-Qian:
upstream ticket in vim repo had been closed, and recently there are some vim window operation code/logic were patched, so supposed now it was ok? or whatever, please expressed your thinking there about if vim itself can make plugin's life easy.. :smile:

Issue #776, opened 2021-06-23
[bug] improper ctags argument ordering


XSven:
This example tagbar debug output

`2021-06-23 07:49:50.071215: ctags_args (is a list): ['--extras=+F', '-f', '-', '--format=2', '--excmd=pattern', '--fields=nksSafet', '--sort=no', '--append=no', '-V', '--options=~/.vim/ctags/perl.ctags', '--language-force=perl', '--perl-kinds=pcMflds']`

shows that a `g:tagbar_ctags_options` setting that leads to `--options=~/.vim/ctags/perl.ctags` comes before `--perl-kinds=pcMflds` . With this order it is not possible to turn of kinds like for example **p** configuring `--kinds-perl=-p` in the _~/.vim/ctags/perl.ctags_ file. I would assume that ctags options specified via a `g:tagbar_ctags_options` should come last and have higher precedence than the ctags options prepared by the tagbar plugin itself.

This is how you can test it. Create a file _Foo.pm_ with the content **package Foo;**. Then run either

```
ctags -o - --kinds-perl=-p --perl-kinds=p Foo.pm
Foo     Foo.pm  /^package Foo;$/;"      p
```

or this

```
ctags -o - --perl-kinds=p --kinds-perl=-p Foo.pm
# no output
```

PR #876, opened 2024-03-08
Allow custom language definitions

### LINKED ISSUE #789: #777 breaks the ability to add a definition for a new language or file type

AndyMcAliley:
Fixes #789 

This PR adds a second duplicate `--option` directive to the `ctags` command, using `deffile` twice. The first `--option` comes prior to `--language-force` and lets ctags know the custom language. The second `--option` comes last to override the default `--kinds-<lang>` definition.

This PR addresses the same problem that #792 addresses, but in a different way that doesn't require additional user options.

This fix needs testing, especially on the use case described in #776.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -232,18 +232,18 @@ endfunction
 function! s:InitTypes() abort
     call tagbar#debug#log('Initializing types')
 
-    let supported_types = s:GetSupportedFiletypes()
+    let s:supported_types = s:GetSupportedFiletypes()
 
     if s:ctags_is_uctags
-        let s:known_types = tagbar#types#uctags#init(supported_types)
+        let s:known_types = tagbar#types#uctags#init(s:supported_types)
     else
-        let s:known_types = tagbar#types#ctags#init(supported_types)
+        let s:known_types = tagbar#types#ctags#init(s:supported_types)
     endif
 
     " Use dart_ctags if available
     let dart_ctags = s:CheckFTCtags('dart_ctags', 'dart')
     if dart_ctags !=# ''
-        let supported_types['dart'] = 1
+        let s:supported_types['dart'] = 1
         call tagbar#debug#log('Detected dart_ctags, overriding typedef')
         let type_dart = tagbar#prototypes#typeinfo#new()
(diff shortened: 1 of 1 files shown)


AndyMcAliley:
Actually, maybe it would be even better to check if the language is recognized by ctags, and only include the first `--option` directive for custom languages.

Arhinebeneze1 (approved):

raven42 (approved):
Looks like this works correctly for both the original failing case #776 and the new language case #789

Thanks for the contribution!

PR #893, opened 2024-10-24
This adds support for controlling display of suffix/prefix

### LINKED ISSUE #870: Markdown: How to hide scopes?
Seek for help: 
I want to use tagbar with markdown, but do not display the scope info of each title like following:

image

How to do to disable that display?

raven42:
Fixes #870

This adds the `g:tagbar_show_prefix` and `g:tagbar_show_suffix` options. These will help control if the prefix info and/or suffix info is printed in the tagbar window for the tag.

This is just test code for now.


raven42:
Finally got a chance to test this out. The new options now work as expected.

PR #895, opened 2024-11-09
Higher tolerance for splitting tag line

roccomao:
This is a very small patch, use `;"\t` to split tags line instead of `;"`. Using `;"` to split tags line will cause tagbar to ignore those tags which `;"` in the current line. It is indeed possible to have `;"` in a line of a file, such as a string `";"`, or a similar string ending with `;` such as html entity (like `"&semi;"`), or included in the end-of-line _comments_, etc. But the probability of `;"\t` appearing in a code file is much smaller. In addition, the patch fix a typo that was seen.

### DIFF
--- a/autoload/tagbar.vim
+++ b/autoload/tagbar.vim
@@ -1319,7 +1319,7 @@ function! s:ProcessFile(fname, ftype) abort
 
         let seen[line] = 1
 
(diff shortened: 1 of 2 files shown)


raven42:
Looks good. Thanks for the contribution!

